PR #6, opened 2020-06-18
Implement basic dispatch function and exec loop.

whiten:
The hard part here was creating a map of connections by database name
that's persistent for the life of the library. This HashMap can't be
held at the file level because JsValue doesn't implement the Send trait,
and so the compiler has to prove it is never shared between threads.

replicache_client.js: 13437
replicache_client.js.br: 3180
replicache_client_bg.wasm: 52126
replicache_client_bg.wasm.br: 22230


src/dispatch.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+struct Dispatcher {
+    connections: HashMap<String, IdbStore>,
+}
+
+impl Dispatcher {
+    async fn open(&mut self, req: &Request) -> Response {
+        if req.db_name.len() == 0 {
```
aboodman, line 51:
```suggestion
        if req.db_name.is_empty() {
```

src/dispatch.rs
```diff
@@ -0,0 +1,103 @@
...
+pub async fn dispatch(db_name: String, rpc: String, data: String) -> Response {
+    let (tx, rx) = channel::<Response>(1);
+    let request = Request {
+        db_name,
+        rpc,
+        data,
+        response: tx,
+    };
+    {
+        SENDER.lock().unwrap().send(request).await;
```
aboodman, line 97: Avoid `unwrap`. Instead log the error?

src/dispatch.rs
whiten: Done.

src/dispatch.rs
whiten: In this case the Result is only to protect against the mutex having been "poisoned" (see https://doc.rust-lang.org/std/sync/struct.Mutex.html), meaning that a holder of it panicked. You can test for this and access the value without acquiring the mutex, but I think at that point the error is unrecoverable, and there's no way to put the mutex back in a good state.

In such cases, do you think we should go through the trouble to trying to explicitly report this (the triggering panic should have already been reported) or just leave this be? My vote is to just `unwrap()` here, but I agree that it's too bad that these semantics make code reviewing unwrap() calls tricky since it's sometimes verbotten and sometimes boilerplate.

whiten:
> Do you know what's going on with the size increase? It's up over 3x, and that doesn't seem to correspond to the source code increase.
> 
> Otherwise this looks great!

whiten:
> Do you know what's going on with the size increase? It's up over 3x, and that doesn't seem to correspond to the source code increase.
> 
> Otherwise this looks great!

So this goes from 15272 to 50256, or a difference of 34,984 bytes uncompressed. Using the build with symbols from https://github.com/rocicorp/repc/pull/8 and 'twiggy diff', I've grouped the changes as:
o 6927: fmt
o 5144: data[0]
o 3557: hash
o 2379: alloc
o 1798: core::str/unicode
o 832: sync::once
o 14347: <other>

"hash" and "sync::once" changes are obvious, those are new facilities I'm using. "fmt" also grows in https://github.com/rocicorp/repc/pull/7 (almost certainly overlapping with here), and probably comes from the first uses of Result.unwrap()/expect() and log!(), both of which embed format!() calls. I imagine some of the alloc and core::str/core::unicode stuff come in as a result of format!() too.

I played with this for too long, and we're in this weird period where we're bringing common stuff in for the first time, so tiny changes (e.g., using e.to_string() on an error) result in large swings in binary size. My instinct is that it's best for us to largely ignore this for now and do the equivalent of profile-driven optimization once we've stuff the basic stack up, as right now 5% "savings" can be easily achieved but wont be defensible at all as we keep adding more error handling and so on.

aboodman:
On Mon, Jun 22, 2020 at 12:32 PM Nate White <notifications@github.com>
wrote:

> Do you know what's going on with the size increase? It's up over 3x, and
> that doesn't seem to correspond to the source code increase.
>
> Otherwise this looks great!
>
> So this goes from 15272 to 50256, or a difference of 34,984 bytes
> uncompressed. Using the build with symbols from #8
> <https://github.com/rocicorp/repc/pull/8> and 'twiggy diff', I've grouped
> the changes as:
> o 6927: fmt
> o 5144: data[0]
> o 3557: hash
> o 2379: alloc
> o 1798: core::str/unicode
> o 832: sync::once
> o 14347:
>
> "hash" and "sync::once" changes are obvious, those are new facilities I'm
> using. "fmt" also grows in #7 <https://github.com/rocicorp/repc/pull/7>
> (almost certainly overlapping with here), and probably comes from the first
> uses of Result.unwrap()/expect() and log!(), both of which embed format!()
> calls. I imagine some of the alloc and core::str/core::unicode stuff come
> in as a result of format!() too.
>
I had been trying to avoid pulling in fmt. I'm not sure if that is
practical to do, realistically.

> I played with this for too long, and we're in this weird period where
> we're bringing common stuff in for the first time, so tiny changes (e.g.,
> using e.to_string() on an error) result in large swings in binary size. My
> instinct is that it's best for us to largely ignore this for now and do the
> equivalent of profile-driven optimization once we've stuff the basic stack
> up, as right now 5% "savings" can be easily achieved but wont be defensible
> at all as we keep adding more error handling and so on.
>
I'm OK with that as long as you think that sufficient tooling exists to do
so. I hadn't really dug into it yet, and was so using the "start fast and
don't get slower" approach. Maybe that approach isn't as applicable to size.

src/dispatch.rs
aboodman: We talked about this early on and came to this consensus:

> * If possible, do not intentionally crash processes (eg, by panicing on unexpected conditions).
>    * For our client-side software, this is bad practice because we frequently run in-process with our users and we will crash them too!

-- https://github.com/rocicorp/replicache/blob/master/contributing.md#style-general

I think the core consideration is that we do not ever want to crash our host software.

However. I do not know exactly what panic and crash mean in wasm. I assume (?) that if a wasm program panics the wasm module goes away but the webpage is otherwise unaffected. In that case, perhaps we don't care.

But then we should think about what it means to panic in other environments.

If we can arrange for a global panic handler in Rust to convert panics into logs that don't crash the host program in the environments we care about, then I agree that would be much more convenient than meticulously avoiding panics.

cc @phritz

whiten: I changed this call, which ended up being easy. I just wanted to say that I agree with the sentiment above, but this is still a weird case since Mutex.lock() is defined as not being allowed to error unless another holder of it has already panicked and left the mutex in a permanently poisoned state.

Btw, I did a little research on trying to avoid panics in Rust, and it sounds like it's considered to be a hard problem because they're so baked in to the standard library. I did find a crate (https://crates.io/crates/no-panic/0.1.5) with a macro that forces the compiler to prove that a function wont panic, but it's incompatible with async functions. :( Googling for "Rust no_panic" yields a bunch of interesting reading, however.

PR #7, opened 2020-06-18
Address warnings, and run 'cargo fmt' on everything.

whiten:
Going forward, we can enforce zero warnings either by putting
"#![deny(warnings)]" at the top of src/lib.rs, or running "cargo rustc
-- -D warnings" as part of CI.

replicache_client.js: 12724
replicache_client.js.br: 3068
replicache_client_bg.wasm: 18485
replicache_client_bg.wasm.br: 8599


src/idbstore.rs
```diff
@@ -20,15 +21,17 @@ impl IdbStore {
         let request = factory.open(name)?;
         let (sender, receiver) = oneshot::channel::<()>();
         let callback = Closure::once(move || {
-            sender.send(());
+            sender.send(()).expect("oneshot send failed");
```
aboodman, line 24: Don't crash. Use log! or print! instead.

whiten: Fixed with log!.

aboodman:
LGTM, once you sign the CA I'll set you up so you can land this.

PR #11, opened 2020-06-25
Implement onupgrade and get/put for idbstore.

whiten:
This also starts using Rust error techniques as from
https://www.philipdaniels.com/blog/2019/defining-rust-error-types/ to
allow using the ? operator in idbstore.rs rather than match everywhere.
I expect we can move towards something similar in dispatch.rs as well.

Much of the size increase here came in with the simple uses of Serde
that this introduced. I think it would be worthwhile to explore exposing
a function per RPC request to avoid this, but didn't want to bite that
off before sharing this.

replicache_client.js: 16321 (+21%)
replicache_client.js.br: 3531 (+11%)
replicache_client_bg.wasm: 128657 (+147%)
replicache_client_bg.wasm.br: 56310 (+154%)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,12 +37,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8aac770f1885fd7e387acedd76065302551364496e46b3dd00860b2f8359b9d"
 
 [[package]]
-name = "bitflags"
-version = "1.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
-
-[[package]]
 name = "block-buffer"
 version = "0.7.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


aboodman:
Bummer. Really curious where the size ends up at. 100k? 200k? 500k?

aboodman:
But otherwise, this looks beautiful on first scan, looking forward to reading through.

whiten:
I just checked, and if I comment out *all* of get(), put(), and the onupgradeneeded callback in idbstore.rs as well as the debug format of JsValue as part of the `impl From<JsValue> for StorageError`, replicache_client_bg.wasm still increases to 124617, so all of that is only contributing 3.2% on top of the rest of the diff.

Serde and serde_json can't know that we're only trying to pass strings through, and thus unconditionally brings in float parsing and formatting, which I think is worth 12k on it's own. Overall, if I run `twiggy diff ../repc2/pkg/replicache_client_bg.wasm.debug pkg/replicache_client_bg.wasm.debug --regex <pattern>` (where `../repc2` contains the previous head), I get the following for various patterns:

serde +19343
data\[ +18743 (not sure what this is)
fmt +17312 (largely overlapping with float, next)
float +12542

PR #12, opened 2020-06-25
Proof-of-concept hiding kv impl behind async trait.

whiten:
replicache_client.js: 16321
replicache_client.js.br: 3531
replicache_client_bg.wasm: 129043
replicache_client_bg.wasm.br: 56334

### DIFF
--- a/src/dispatch.rs
+++ b/src/dispatch.rs
@@ -1,4 +1,5 @@
 use crate::kv::idbstore::IdbStore;
+use crate::kv::Store;
 use async_std::sync::{channel, Receiver, Sender};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
@@ -53,8 +54,8 @@ async fn dispatch_loop(rx: Receiver<Request>) {
                     }
                 };
                 let response = match req.rpc.as_str() {
-                    "get" => dispatcher.get(db, &req.data).await,
-                    "put" => dispatcher.put(db, &req.data).await,
+                    "get" => dispatcher.get(&**db, &req.data).await,
+                    "put" => dispatcher.put(&**db, &req.data).await,
                     _ => Err("Unsupported rpc name".to_string()),
                 };
                 req.response.send(response).await;
@@ -83,7 +84,7 @@ struct PutRequest {
 }
 
 struct Dispatcher {
-    connections: HashMap<String, IdbStore>,
+    connections: HashMap<String, Box<dyn Store>>,
 }
 
 impl Dispatcher {
@@ -101,7 +102,7 @@ impl Dispatcher {
             }
             Ok(v) => {
                 if let Some(v) = v {
-                    self.connections.insert(req.db_name.clone(), v);
+                    self.connections.insert(req.db_name.clone(), Box::new(v));
                 }
             }
         }
@@ -117,7 +118,7 @@ impl Dispatcher {
         Ok("".to_string())
     }
 
-    async fn get(&self, db: &IdbStore, data: &String) -> Response {
+    async fn get(&self, db: &dyn Store, data: &String) -> Response {
         let req: GetRequest = match serde_json::from_str(data) {
             Ok(v) => v,
             Err(_) => return Err("Failed to parse request".into()),
@@ -132,7 +133,7 @@ impl Dispatcher {
         }
     }
 
-    async fn put(&self, db: &IdbStore, data: &String) -> Response {
+    async fn put(&self, db: &dyn Store, data: &String) -> Response {
         let req: PutRequest = match serde_json::from_str(data) {
             Ok(v) => v,
             Err(_) => return Err("Failed to parse request".into()),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,6 +31,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c17772156ef2829aadc587461c7753af20b7e8db1529bc66855add962a3b35d3"
 
 [[package]]
+name = "async-trait"
+version = "0.1.36"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a265e3abeffdce30b2e26b7a11b222fe37c6067404001b434101457d0385eb92"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -457,6 +468,7 @@ name = "replicache-client"
 version = "0.1.0"
 dependencies = [
  "async-std",
+ "async-trait",
  "console_error_panic_hook",
  "data-encoding",
  "futures",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ default = ["console_error_panic_hook"]
 
 [dependencies]
 async-std = { version = "=1.6.0", features = ["unstable"] }
+async-trait = "0.1.36"
 console_error_panic_hook = { version = "0.1.1", optional = true }
 data-encoding = "1.1.1"
 futures = "0.3.5"
--- a/src/kv/idbstore.rs
+++ b/src/kv/idbstore.rs
@@ -1,33 +1,30 @@
+use crate::kv::{Store, StoreError};
+use async_trait::async_trait;
 use futures::channel::oneshot;
 use std::fmt;
 use wasm_bindgen::closure::Closure;
 use wasm_bindgen::{JsCast, JsValue};
 use web_sys::IdbDatabase;
 
-#[derive(Debug)]
-pub enum StorageError {
-    Str(String),
-}
+type Result<T> = std::result::Result<T, StoreError>;
 
-type Result<T> = std::result::Result<T, StorageError>;
-
-impl From<JsValue> for StorageError {
-    fn from(err: JsValue) -> StorageError {
+impl From<JsValue> for StoreError {
+    fn from(err: JsValue) -> StoreError {
         // TODO(nate): Pick out a useful subset of this value.
-        StorageError::Str(format!("{:?}", err))
+        StoreError::Str(format!("{:?}", err))
     }
 }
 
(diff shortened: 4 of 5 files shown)


src/dispatch.rs
```diff
@@ -83,7 +84,7 @@ struct PutRequest {
 }
 
 struct Dispatcher {
-    connections: HashMap<String, IdbStore>,
+    connections: HashMap<String, Box<dyn Store>>,
```
aboodman, line 87: It seems like dynamic dispatch is vaguely discouraged in Rust, so I was trying in that thing I sent in Slack to find a way to not use it*.

I do not think that "idiomatic Rust" should be a goal for this project right now, but at the same time it's interesting to explore tradeoffs to learn the language.

Anyway, two questions:

1. Does async-trait end up compiling a `dyn Trait` result?
2. In TypeScript, `async foo() -> Bar` is another way to say `Foo() -> Promise<Bar>`. The functions both end up with the same type (the second one). The only difference is that inside the first you can use `await`. I wonder if it's the same in Rust. Can we avoid async-trait (and thereby have option to use static dispatch) by phrasing the trait with return type Future, but still *implement* the trait using async/await.

* OTOH, static dispatch is associated with increased code size which is the opposite of what we want **.

** OTOOH, in our case we'll only have one implementation compiled into the binary at a time so maybe we wouldn't see a size increase.

whiten: > It seems like dynamic dispatch is vaguely discouraged in Rust, so I was trying in that thing I sent in Slack to find a way to not use it*.

In a quick reading, I do see folks talking about not using it, but just to avoid the small overhead that calling through a C++ vtable introduces. That is, I don't think that's it's discouraged b/c of Rust, but just because you *can* use generics to squeak out that extra .00001% of performance if you wish. Is there a further reason I missed?

> I do not think that "idiomatic Rust" should be a goal for this project right now, but at the same time it's interesting to explore tradeoffs to learn the language.
> 
> Anyway, two questions:
> 
> 1. Does async-trait end up compiling a `dyn Trait` result?

https://docs.rs/async-trait/0.1.36/async_trait/#explanation suggests that methods in async traits return a `Pin<Box<dyn Future + Send + 'async>>` (I opted out of the `Send` part via `#[async_trait(?Send)]` since IdbDatabase doens't support that).

I don't know if that answers your question though, does it? I'm cargo culting far more than I'm comfortable with because I'm not grokking much of what's going on yet. :/

> 2. In TypeScript, `async foo() -> Bar` is another way to say `Foo() -> Promise<Bar>`. The functions both end up with the same type (the second one). The only difference is that inside the first you can use `await`. I wonder if it's the same in Rust. Can we avoid async-trait (and thereby have option to use static dispatch) by phrasing the trait with return type Future, but still _implement_ the trait using async/await.

I don't understand yet how using async-trait prevents us from doing static dispatch here. I think that we *could* define `dispatch_loop` and `Dispatcher` as generics on a type implementing a trait, and then instantiate that once in each of our build products based on the Store implementation we want. Does the `dyn Future` in async-trait preclude that? In either case, especially at a level so close to mapping to I/O, I don't think a bit of indirection overhead will ever show up in a profile unlike how it might in some big in-memory graph algorithm.

> * OTOH, static dispatch is associated with increased code size which is the opposite of what we want **.
> 
> ** OTOOH, in our case we'll only have one implementation compiled into the binary at a time so maybe we wouldn't see a size increase.

aboodman:
On Thu, Jun 25, 2020 at 10:05 AM Nate White <notifications@github.com>
wrote:

> *@whiten* commented on this pull request.
>
> In src/dispatch.rs
> <https://github.com/rocicorp/repc/pull/12#discussion_r445807625>:
>
> > @@ -83,7 +84,7 @@ struct PutRequest {
>  }
>
>  struct Dispatcher {
> -    connections: HashMap<String, IdbStore>,
> +    connections: HashMap<String, Box<dyn Store>>,
>
> It seems like dynamic dispatch is vaguely discouraged in Rust, so I was
> trying in that thing I sent in Slack to find a way to not use it*.
>
> In a quick reading, I do see folks talking about not using it, but just to
> avoid the small overhead that calling through a C++ vtable introduces. That
> is, I don't think that's it's discouraged b/c of Rust, but just because you
> *can* use generics to squeak out that extra .00001% of performance if you
> wish. Is there a further reason I missed?
>
> I do not think that "idiomatic Rust" should be a goal for this project
> right now, but at the same time it's interesting to explore tradeoffs to
> learn the language.
>
> Anyway, two questions:
>
>    1. Does async-trait end up compiling a dyn Trait result?
>
> https://docs.rs/async-trait/0.1.36/async_trait/#explanation suggests that
> methods in async traits return a Pin<Box<dyn Future + Send + 'async>> (I
> opted out of the Send part via #[async_trait(?Send)] since IdbDatabase
> doens't support that).
>
> I don't know if that answers your question though, does it? I'm cargo
> culting far more than I'm comfortable with because I'm not grokking much of
> what's going on yet. :/
>
Me too.

>
>    1. In TypeScript, async foo() -> Bar is another way to say Foo() ->
>    Promise<Bar>. The functions both end up with the same type (the second
>    one). The only difference is that inside the first you can use await.
>    I wonder if it's the same in Rust. Can we avoid async-trait (and thereby
>    have option to use static dispatch) by phrasing the trait with return type
>    Future, but still *implement* the trait using async/await.
>
> I don't understand yet how using async-trait prevents us from doing static
> dispatch here.
>
From
https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/,
async-trait generates the return type:

Pin<Box<dyn Future<Output = T> + Send + '_>>

Rather than what async functions normally generate:

impl Future<Output = T> + '_

The `dyn` causes dynamic dispatch.

> Does the dyn Future in async-trait preclude that?
>
I think so.

> In either case, especially at a level so close to mapping to I/O, I don't
> think a bit of indirection overhead will ever show up in a profile unlike
> how it might in some big in-memory graph algorithm.
>
Agree completely. Perf not a thing to care about here.

I guess the other reason I'm iffy on async-trait is just a desire to use as
little Rust as possible, while we're learning it. Like I'd just avoid
traits completely if I could, and even Futures, just to be able to deeply
understand all the code we're using.

Along those lines, I'm again considering not defining a trait for dagstore
at all and just linking directly against idbstore (and doing all the tests
as integration tests).

WDYT about just putting aside the question of how to define this trait for
now and coming back to it when dagstore starts coming together? Maybe
writing integration tests will be fine and avoid this whole topic.

whiten:
> I guess the other reason I'm iffy on async-trait is just a desire to use as
little Rust as possible, while we're learning it. Like I'd just avoid
traits completely if I could, and even Futures, just to be able to deeply
understand all the code we're using.

I'm pretty pessimistic on this. The language feels like a beast, and things like error handling feel burdensome until I used the custom-error-enum-with-from thing at the top of idbstore to enable the ? operator. I'm just not getting the impression that there's a simpler subset of Rust to use that avoids tough cases. Maybe my problem is finding the right documentation to read, but every blog seems to be pulling in its favorite crates.

I think we're forced into using futures because WASM is single-threaded, so that's not avoidable. The Send, Display, and Debug traits have already all forced themselves on me through normal work, so I sort of expect them to keep creeping in.

I think I'm maybe more anxious to cordon off the WASM stuff and be able to just use `cargo test` for much of what we're going to write soon to get closer to a normal Rust experience rather than the weird wasm-bindgen no-unit-tests world we're in now.

> WDYT about just putting aside the question of how to define this trait for now and coming back to it when dagstore starts coming together? Maybe writing integration tests will be fine and avoid this whole topic.

Integration tests only avoid this topic until we care for this to work on mobile. I would personally rather put idbstore behind a trait interface early and then mock that with a non-JS implementation so we can put the right stuff into unit tests and the right stuff into integration tests as we build. One of Rust's benefits is supposed to be making refactoring safer, so I expect that if we find we don't like our first pass at that interface we'll be able to revisit.

aboodman:
On Thu, Jun 25, 2020 at 11:39 AM Nate White <notifications@github.com>
wrote:

> I guess the other reason I'm iffy on async-trait is just a desire to use as
> little Rust as possible, while we're learning it. Like I'd just avoid
> traits completely if I could, and even Futures, just to be able to deeply
> understand all the code we're using.
>
> I'm pretty pessimistic on this. The language feels like a beast, and
> things like error handling feel burdensome until I used the
> custom-error-enum-with-from thing at the top of idbstore to enable the ?
> operator. I'm just not getting the impression that there's a simpler subset
> of Rust to use that avoids tough cases. Maybe my problem is finding the
> right documentation to read, but every blog seems to be pulling in its
> favorite crates.
>
I mean there are little things. Like I don't think we *have* to use async,
nor async-trait.

> I think we're forced into using futures because WASM is single-threaded,
> so that's not avoidable. The Send, Display, and Debug traits have already
> all forced themselves on me through normal work, so I sort of expect them
> to keep creeping in.
>
Sure I should have been more clear. We'll have to use traits -- I was
wondering about avoiding defining our own.

> I think I'm maybe more anxious to cordon off the WASM stuff and be able to
> just use cargo test for much of what we're going to write soon to get
> closer to a normal Rust experience rather than the weird wasm-bindgen
> no-unit-tests world we're in now.
>
> WDYT about just putting aside the question of how to define this trait for
> now and coming back to it when dagstore starts coming together? Maybe
> writing integration tests will be fine and avoid this whole topic.
>
> Integration tests only avoid this topic until we care for this to work on
> mobile.
>
Right, but that's a long ways off.

> I would personally rather put idbstore behind a trait interface early and
> then mock that with a non-JS implementation so we can put the right stuff
> into unit tests and the right stuff into integration tests as we build. One
> of Rust's benefits is supposed to be making refactoring safer, so I expect
> that if we find we don't like our first pass at that interface we'll be
> able to revisit.
>
One downside of doing that, though, is that our tests will tend to not
exercise the real implementation of the kvstore. The one that our customers
will use in real life... In Go this would not bother me as much but when we
are new to a language and there are more sharp edges, I'm more concerned
about missing something by focusing our tests on a mock impl.

For some reason the weird wasm tests don't bother me much yet.

Hm.

PR #13, opened 2020-06-26
Use 'log' crate as a logging facade w/ console_log.

whiten:
This removes the WASM-only dependency on web_sys::console from most of
our sources, instead registering the 'console_log' crate as the log
emitter in src/wasm.rs. When we have non-WASM targets, we can register a
different logger for them from the list at
https://docs.rs/log/0.4.8/log/#available-logging-implementations.

replicache_client.js: 16761 (+2.7%)
replicache_client.js.br: 3613 (+2.4%)
replicache_client_bg.wasm: 131254 (+2%)
replicache_client_bg.wasm.br: 57200 (+1.4%)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -137,6 +137,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "console_log"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "501a375961cef1a0d44767200e66e4a559283097e91d0730b1d75dfb2f8a1494"
+dependencies = [
+ "log",
+ "web-sys",
+]
+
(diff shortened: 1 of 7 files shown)


whiten:
An annoying thing this uncovered is that wasm-bindgen-test-runner in Chrome headless mode captures console.log() and console.error() output, but *not* console.info() or console.warning(). I could not find any references to this, and the closest I got to a code explanation is https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs#L161-L163, but I'm not very familiar with browser testing so don't have good instincts to get to the bottom of it. All log levels are present when testing without --headless, however.

PR #14, opened 2020-06-26
Switch from serde to nanoserde to reduce deps.

whiten:
We'll learn over time if this is feature-rich enough for us, but since
the JSON we're passing here is for a Replicache-internal API, I suspect
it may be.

replicache_client.js: 16321 (~=)
replicache_client.js.br: 3537 (~=)
replicache_client_bg.wasm: 106621 (-17%)
replicache_client_bg.wasm.br: 45153 (-20%)


whiten:
I'm unsure of this given that we're going to have JSON communication between repc and the server as well. Undecided as to whether we should switch to this and see if we can work through it's deficiencies, or stay with the broadly used serde crate. Good to know it exists, in any case.

aboodman:
aw yeah.

PR #18, opened 2020-06-27
Add MemStore implementation of kv::Store

aboodman:
This required changing Store::put() to borrow self mutably. Followed it up the tree and feel I have it right, but borrow checker disagrees:

```
rror[E0502]: cannot borrow `dispatcher` as immutable because it is also borrowed as mutable
  --> src/dispatch.rs:56:30
   |
46 |                     match dispatcher.connections.get_mut(&req.db_name[..]) {
   |                           ---------------------- mutable borrow occurs here
...
56 |                     "has" => dispatcher.has(&**db, &req.data).await,
   |                              ^^^^^^^^^^     ----- mutable borrow later used here
   |                              |
   |                              immutable borrow occurs here
```

The way I see it, dispatcher is borrowed mutably to call `get_mut` but that reference is immediately dropped when the `let db = {...}` scope exits. For some reason borrow checker sees the mutable ref as living on, same lifetime as the db reference.

### DIFF
--- a/src/kv/memstore.rs
+++ b/src/kv/memstore.rs
@@ -1,7 +1,6 @@
-use async_trait::async_trait;
 use crate::kv::{Store, StoreError};
+use async_trait::async_trait;
 use std::collections::HashMap;
-use std::fmt;
 
 type Result<T> = std::result::Result<T, StoreError>;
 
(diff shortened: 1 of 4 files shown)


src/kv/memstore.rs
```diff
@@ -0,0 +1,55 @@
...
+    use futures::executor::block_on;
+
+    #[test]
+    fn basics() {
+        let mut ms = MemStore::new();
+        assert_eq!(false, block_on(ms.has("foo")).unwrap());
+        assert_eq!(None, block_on(ms.get("foo")).unwrap());
+        block_on(ms.put("foo", "bar".as_bytes())).unwrap();
+        assert_eq!(true, block_on(ms.has("foo")).unwrap());
+        assert_eq!(Some("bar".as_bytes().to_vec()),
```
whiten, line 50: I still plan to set up for GitHub workflow checks, but you should turn on the rustfmt auto-formatter in your editor.

PR #19, opened 2020-06-27
Memstore

whiten:


src/kv/memstore.rs
```diff
@@ -0,0 +1,37 @@
...
+        Ok(Some(MemStore {
+            map: HashMap::new(),
+        }))
+    }
+}
+
+#[async_trait(?Send)]
+impl Store for MemStore {
+    async fn put(&mut self, key: &str, value: &[u8]) -> Result<()> {
+        self.map.insert(key.to_string(), value.to_vec());
```
arv, line 23: Why to string here but not below?

aboodman: The parameter types are different.

Compare https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert vs https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.contains_key.

insert() needs to take ownership of the key because its going to store it in the map. So you have to pass it an owned string, aka `String`, which you can get from `&str` with `to_string()`, among other ways.

the other two methods don't need to take ownership of the key, they're just comparing keys they already have to the specified one. So you can pass a 'borrowed' string, aka &str.

aboodman: (to_vec()) is doing the same thing to the `&[u8]` reference -> copying it into an owned vector that the hashmap can take ownership of.

aboodman: (ps @whiten maybe this means that our put method should take `Vec` since the caller probably already has an owned copy, no sense copying it again).

aboodman: > maybe this means that our put method should take Vec since the caller probably already has an owned copy, no sense copying it again

Nevermind. I remember it is as it is now on purpose. In the persistent storage case, my hope was that eventually we can take the `&[u8]` all the way to where it is streamed into whatever the persistent storage is without any extra copies. (Right now we copy it again into the JS buffer type blech, but passing an owned vector wouldn't help one way or other with that).

PR #20, opened 2020-06-29
First pass at transactional interface for Store.

whiten:
Only implemented for MemStore so far, IdbStore returns a "Not supported"
Err on both .read() and .write().


src/kv/memstore.rs
```diff
@@ -1,55 +1,180 @@
...
+    async fn put(&self, key: &str, value: &[u8]) -> Result<()> {
+        self.check_open()?;
+        self.pending
+            .lock()
+            .await
+            .insert(key.to_string(), value.to_vec());
+        Ok(())
+    }
+
+    async fn commit(&mut self) -> Result<()> {
```
aboodman, line 125: Did you try having `commit` take `self` by value? In some iterations of my attempt, this was a nice way to prevent `commit` from getting called twice without needing `check_open`.

src/kv/memstore.rs
```diff
@@ -1,55 +1,181 @@
...
+        match self.store.map.lock().await.get(key) {
             None => Ok(None),
             Some(v) => Ok(Some(v.to_vec())),
         }
     }
 }
+
+struct WriteTransaction<'a> {
+    rt: ReadTransaction<'a>,
+    pending: Mutex<HashMap<String, Vec<u8>>>,
```
aboodman, line 75: This will have to get slightly fancier once del is implemented.

src/kv/memstore.rs
whiten: No, but good instinct, and it's nice that it proves that the transaction isn't linked in anywhere either. Will try that out soon.

src/kv/memstore.rs
whiten: Yep!

PR #21, opened 2020-07-02
Chunk (de)serialization.

aboodman:
Chunk (de)serialization.


src/dag/chunk.rs
```diff
@@ -1,8 +1,100 @@
...
+
+    pub fn hash(&self) -> &str {
+        &self.hash
+    }
+
+    pub fn data(&self) -> &[u8] {
+        &self.data
+    }
+
+    pub fn refs<'a>(&'a self) -> Box<dyn Iterator<Item = &str> + 'a> {
```
aboodman, line 39: I was very sad to return the boxed iterator, but I wanted to flatten then case of no meta buffer (as will be the case for leaf nodes of the dag) to an empty iterator so that callers don't need to worry about. In that case, the return iterator can either be the generated one from FlatBuffers or std::iter::Empty.

I couldn't find any other way to flatten these cases... If the FlatBuffers had a way to get out &[&str] that would work, but I don't think it does.

src/dag/chunk.rs
```diff
@@ -1,8 +1,100 @@
...
+        match self.meta() {
```
aboodman, line 40: It seems like you ought to be able to say something like:

```
meta_generated::meta::get_root_as_meta(self.meta()?).refs()?.or(std::iter::empty())
```

... but could not figure that out.

src/dag/chunk.rs
```diff
@@ -1,8 +1,100 @@
...
+            Some((buf, offset)) => Some(&buf[*offset..]),
+        }
+    }
+
+    fn create_meta(refs: &[&str]) -> Option<(Vec<u8>, usize)> {
+        if refs.is_empty() {
+            return None;
+        }
+        let mut builder = FlatBufferBuilder::default();
+        let refs = builder.create_vector_of_strings(refs);
```
aboodman, line 61: This allocates, but I'd already given up on the write path by that point.

aboodman:
@whiten please review seriously. I don't yet feel I'm programming in Rust, more like bashing my head against the compiler.

src/dag/meta.fbs
```diff
@@ -0,0 +1,9 @@
+namespace meta;
+
+// Metadata about a chunk, stored separately to enable exploring
```
whiten, line 3: Can you modify this comment to explain what the refs are?

src/dag/key.rs
```diff
@@ -92,10 +90,10 @@ mod tests {
     #[test]
     fn roundtrip() -> Result<(), ParseError> {
         let cases: &[Key] = &[
-            Key::ChunkData("".to_string()),
-            Key::ChunkData("a".to_string()),
-            Key::ChunkRefs("".to_string()),
-            Key::ChunkRefs("a".to_string()),
+            Key::ChunkData(String::from("".to_string())),
```
whiten, line 93: It's a bit annoying to me that there are so many ways to do this conversion! I've taken to doing `"".into()` just because it's the shortest.

Should we perhaps try to standardize on a preferred way to do this str-constant -> String conversion throughout, or not worry about it?

src/dag/chunk.rs
whiten: Should we start adding TODO comments for places like this where we suspect there's a better way and we just don't want to block on finding it right now? I think that may be appropriate.

src/dag/chunk.rs
```diff
@@ -1,8 +1,100 @@
...
 pub struct Chunk {
-    pub key: Key,
-    buf: Vec<u8>,
-    refs: [Key],
+    hash: String,
+    data: Vec<u8>,
+    meta: Option<(Vec<u8>, usize)>,
+}
+
+#[allow(dead_code)]
```
whiten, line 13: I think you can remove this and the `allow` on line 5 now.

src/dag/chunk.rs
whiten: Yeah, I agree that this is annoying; I'd at least like to avoid the alloc for the empty case, but another problem is that the caller is taking ownership of whatever you return, and so you can't just return a static lifetime empty iterator there.

I'd just leave another TODO for now. Another way I would try attacking this later is that if we only expect callers to call this if they plan to consume the whole ref list, you can have them pass in a `mut Vec<&str>`, and copy any refs into it after clearing it, allowing the Vector to be on the caller's stack frame.

src/dag/chunk.rs
```diff
@@ -1,8 +1,100 @@
-pub type Key = String;
+use super::meta_generated;
```
whiten, line 1: Maybe add `::meta` to the end so we can shorten the various `meta_generated::meta` expressions in this file?

src/dag/chunk.rs
whiten: I can barely read this, so I don't know if I'd favor the above!

`?`for `Options`s only works if the return type of the function is itself an `Option`, so not applicable here.

src/dag/chunk.rs
```diff
@@ -1,8 +1,100 @@
...
-    pub key: Key,
-    buf: Vec<u8>,
-    refs: [Key],
+    hash: String,
+    data: Vec<u8>,
+    meta: Option<(Vec<u8>, usize)>,
+}
+
+#[allow(dead_code)]
+impl Chunk {
```
whiten, line 14: This might be an alternate place to put a ~3-line description of what a Chunk logically represents and how `refs` relates to it. It would help me. :)

src/dag/meta_generated.rs
```diff
@@ -0,0 +1,118 @@
+// automatically generated by the FlatBuffers compiler, do not modify
```
whiten, line 1: Only tangentially related to this, I've been debating introducing a `Makefile` in the root to make the various lint and testing commands I'm using easier to access, and that would also be a good place to store whatever invocation creates this.

Are you open to that, opposed, or would you want more of an argument before deciding?

src/dag/chunk.rs
aboodman: Sure but then I might as well just return a new `Vec<String>` or whatever, right? What is flatbuffers really buying us over pb at that point?

aboodman: > another problem is that the caller is taking ownership of whatever you return,

I don't understand this comment. I tried to make it so that callers *is not* taking ownership of returned data (flatbuffer still owns it).

aboodman: I just decided to have it return an option instead. Still added TODO.

src/dag/chunk.rs
aboodman: Yeah I just mean that conceptually it would be nice if you could use the question mark iterator on a chain to get an Option, then at the end convert that to a Result.

src/dag/chunk.rs
aboodman: Done.

src/dag/meta.fbs
aboodman: Done.

src/dag/key.rs
aboodman: Yeah I hate how many ways there are to convert to and fro in Rust -- or at least that's my perception so far, not knowing what they all do. Happy to standardize on into().

src/dag/chunk.rs
aboodman: done

src/dag/chunk.rs
aboodman: done

src/dag/chunk.rs
aboodman: I added some comments to the module explaining the overall approach, plus here, let me know if it works for you.

src/dag/meta_generated.rs
aboodman: In favor of the idea, but never took the time to learn Makefile syntax. Can we use bash instead? (I think actually we should probably bazelfy this at some point, but...)

Issue #3, opened 2020-06-17
RFC: Overall approach


aboodman:
# repc — from keys to values and back again.

This bug describes the high-level architectural approach to rewriting replicache-client in Rust. Note that all the examples are in psuedo-Go because all eyes on this still know Go better 😂.

## Layercake

```
+---------------------------------------------------------------------+
| RPC       | Simple, stateless cross-lang interface (port of `repm`) |
+---------------------------------------------------------------------|
| DB        | Top-level Replicache interface (port of Go `DB`)        |
+---------------------------------------------------------------------|
| ProllyMap | Implements map prolly tree atop DAG                     |
+---------------------------------------------------------------------|
| DAG       | Content-addressed DAG store with ref-counting           |
+---------------------------------------------------------------------|
| KV        | Abstracts underlying storage requirements               |
+---------------------------------------------------------------------+
```

## RPC

RPC implements a straight port of the existing `repm` package from Go.

I think it can/should even be bytewise compatible for now. We might want to make this better in the future, but let's cut scope and do it the same way for now.

## DB

This is a straight port of `db.DB` from Go. It doesn't need to have exact same interface, because its only purpose is to support `RPC`. But I don't think we need to think too hard about it -- what we have in Go seems to work well.

Each DB instance (e.g., each Replicache *client*) will own a DAGStore.

(*Note:* There's a tradeoff here -- DAGStore writes are serialized, so if we give each DB one, they can write in parallel, but they won't share chunks. In practice I don't really think either of these considerations matter all that much, and neither way seems easier, so vague pref for performance 🤷‍♂️?)

We can basically port the exact DAG we store in Noms to this, except that we don't have to use wonky Noms `Value` types -- we can encode the dag however we like. I expect it makes sense to use FlatBuffers or something like it. At least in Noms, we found that eliminating copies during decoding were one of our biggest perf wins.

### Implementation Sketch

(error handling omitted for brevity)

```
type Commit struct {
  Name string
  Args json.RawMessage
  Parent Hash
  Value Hash
}

func ExampleIncrementTransaction(key string, delta int) {
    kv := NewIDBStore()
    cs := NewDAGStore(kv)
    tx := cs.Write()

    // Get the existing commit
    headHash, _ := tx.GetHead([]byte("main"))
    headChunk, _, _ = tx.Get(headHash)
    var commit Commit
    // Note: Using JSON in this example for illustrative purposes.
    // Serialization is a choice of DB. Lower layers don't care.
[...]

PR #22, opened 2020-07-02
Dagstore

aboodman:
duh nuh nuh nuh duh nuh nuh nuh....

DAGSTORE

Toward https://github.com/rocicorp/repc/issues/3

### DIFF
--- a/src/dag/key.rs
+++ b/src/dag/key.rs
@@ -4,43 +4,47 @@ use std::str;
 // KVKey is the key we use to store our dag data in the underlying
 // kvstore.
 #[derive(Debug, PartialEq, Eq)]
-pub enum Key {
-    ChunkData(String),
-    ChunkRefs(String),
-    Head(String),
-}
-
-impl fmt::Display for Key {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            Key::ChunkData(hash) => write!(f, "c/{}/d", hash),
-            Key::ChunkRefs(hash) => write!(f, "c/{}/r", hash),
-            Key::Head(name) => write!(f, "h/{}", name),
-        }
-    }
+pub enum Key<'a> {
+    ChunkData(&'a str),
+    ChunkMeta(&'a str),
+    Head(&'a str),
 }
 
 type ParseError = ();
 
-impl str::FromStr for Key {
-    type Err = ParseError;
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let parts = s.split("/").collect::<Vec<&str>>();
-        if parts.len() >= 2 {
-            let name = parts[1].into();
-            if parts[0] == "c" {
-                if parts.len() == 3 {
-                    if parts[2] == "d" {
-                        return Ok(Key::ChunkData(name));
-                    } else if parts[2] == "r" {
-                        return Ok(Key::ChunkRefs(name));
-                    }
+// TODO: It would be cool to make this implement FromString trait,
+// as that is the convention, and then "foo".parse() would work.
+// But I got lost in lifetime goop.
+impl<'a> Key<'_> {
+    pub fn parse<'b>(s: &'b str) -> Result<Key<'b>, ParseError> {
+        let mut parts = s.split::<'b>("/");
+        let prefix: &str = parts.next().ok_or(())?;
+        let content = parts.next().ok_or(())?;
+        match prefix {
+            "c" => {
+                let suffix = parts.next().ok_or(())?;
+                if parts.next().is_some() {
+                    return Err(());
                 }
-            } else if parts[0] == "h" {
-                return Ok(Key::Head(name));
-            }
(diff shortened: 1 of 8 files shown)


src/dag/key.rs
```diff
@@ -20,29 +21,34 @@ impl fmt::Display for Key {
     }
 }
 
+/*
```
aboodman, line 24: I could not figure out how to get this to compile, blech.

src/dag/store.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#[allow(dead_code)]
+impl Write {
+    pub fn read(&self) -> &Read {
+        &self.read
+    }
+
+    pub async fn put_chunk(&mut self, c: Chunk) -> Result<()> {
+        self.kvw
+            .put(&Key::ChunkData(c.hash()).to_string(), c.data())
```
aboodman, line 69: I guess maybe it would be easier/better to make this key struct into a pair of bare functions or something.

src/dag/read.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub async fn get_chunk(&mut self, hash: &str) -> Result<Option<Chunk>> {
+        get_chunk(self.kvr.as_ref(), hash).await
+    }
+
+    pub async fn get_head(&mut self, name: &str) -> Result<Option<String>> {
+        get_head(self.kvr.as_ref(), name).await
+    }
+}
+
+pub async fn has_chunk(kvr: &dyn kv::Read, hash: &str) -> Result<bool> {
```
aboodman, line 30: These have to be factored out as bare functions because Rust:

https://users.rust-lang.org/t/casting-traitobject-to-super-trait/33524

aboodman: As with so many things Rust, I'm sure there's a way to do this, but it ends up being easier to just not use advanced features.

src/dag/key.rs
```diff
@@ -4,43 +4,47 @@ use std::str;
...
-                        return Ok(Key::ChunkRefs(name));
-                    }
-                }
-            } else if parts[0] == "h" {
-                return Ok(Key::Head(name));
-            }
+impl<'a> fmt::Display for Key<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Key::ChunkData(hash) => write!(f, "c/{}/d", hash),
```
whiten, line 44: Not specific to this commit, but I just wanted to register that I'm a little iffy on these one-letter prefixes. Just something to think about as we keep building out.

src/dag/key.rs
```diff
@@ -4,43 +4,47 @@ use std::str;
 // KVKey is the key we use to store our dag data in the underlying
 // kvstore.
 #[derive(Debug, PartialEq, Eq)]
-pub enum Key {
-    ChunkData(String),
-    ChunkRefs(String),
-    Head(String),
+pub enum Key<'a> {
```
whiten, line 7: Nice work!

arv: I did read the life time section and just wanted to make sure I understand this change.

`String` is heap allocated. `str` is stack allocated. `'a` parameter means that the lifetime of `Key` must be shorter than the arguments to `ChunkData`/`ChunkMeta`/`Head`?

aboodman: It's not a question of where it is allocated, it's a question of ownership. `String` owns its internal storage. Think of it like `scoped_ptr<std::string>`. `str` is a slice -- it's just a reference to some other storage elsewhere. Think of it like `StringPiece` from Chrome. That storage could be heap allocated or stack.

The lifetime parameters here say:

* There is a struct `Key`
* Which is parameterized by the abstract lifetime `'a` (looking at just this code we don't know what that lifetime is, just that it exists)
* Each of the three fields are a reference to a string elsewhere
* Those three fields all have the same lifetime, the lifetime that is passed in when this type is instanciated

In practice, this is used to avoid having to copy the string data into this struct when it is created. Instead, we can just have its fields reference string data that lives elsewhere.

aboodman: (i think -- @whiten want to crosscheck my explanation?? :))

aboodman: (updated -- this is an enum not a struct)

whiten: Ha, thanks for going first here. I think that's all right, and I just want to emphasize your statement that they "all have the same lifetime", there's no inequality here. But, "they all" is referring to the Key struct and the *references* to the str, not the underlying u8 Vec that backs it.

Thus, anyone with access to the Key could take another reference to that data, and the compiler must assert that it's dropped before the Key is. And, in the Key::parse() function, the compiler can allow a Key with references to the &str that was passed in to escape because the Key is defined to last no longer than those references.

It seems that when we name lifetimes we express lifetime equalities, but with the purpose of allowing the compiler to prove lifetime inequalities, e.g. that the underlying str lasts as long as or longer than the Key that references it.

PR #23, opened 2020-07-06
cargo fmt.

whiten:


whiten:
@aboodman, I have my editor set up to make these changes automatically on save, or running `cargo fmt` performs them as well.

In some cases this is just fixing up `use` ordering or misplaced spaces, while in others it's preferring longer lines and so on. I'd prefer to just lean in to rustfmt (though it does have some knobs if we want to try to have it stop merging lines over 80 columns, or whatever). What do you think?

aboodman:
Yeah I'll figure out how to get my editor to do this today and check in a `.rustfmt`. Apologies.

PR #25, opened 2020-07-06
Implement Store::Read/Write traits for IdbStore.

whiten:
o This also switches Store::commit()/rollback() to move the Store in,
  letting the compiler prevent any subsequent uses. The same pattern had
  to be applied to dag as well otherwise it looked like a dag::Write
  still held a reference. Removed some tests that are now meaningless
  and the compiler wouldn't let stand anymore.

o Moved a few 'use' statements at the top of sources into their test
  modules to avoid unused-use warnings.

o In IdbStore::Write, management of callbacks and state got weird. What
  I settled on is registering transaction oncommit/onabort/onerror
  callbacks when the transaction is first started, and having these
  manage a state variable on the WriteTransaction. Through a muxex and
  condvar, commit() and rollback() are able to query this state or wait
  for it to make a transition.

o All application of pending changes is held back until commit() to
  (hopefully!) ensure that no changes are applied and then autocommitted
  unless they all would be. In future commits, I need to cause the
  IdbTransaction to transition to the error state to further refine
  handling here.

o The Results I return on commit() and rollback() may need some work,
  for instance if the user called commit(), we get a commited callback,
  but the transaction is in an error state, should we really fail the
  commit() call? I'm not happy here yet.

o To enable the IdbStore integration tests, I tried to export a
  new_idbstore() function from wasm.rs, but only for testing. However,
  trying to limit to cfg(test) doesn't work (doctests?), so for now it's
  using cfg(not(default)). I don't even know what that means, but it
  compiles with no warnings.

### DIFF
--- a/src/dag/write.rs
+++ b/src/dag/write.rs
@@ -2,7 +2,6 @@ use super::chunk::Chunk;
 use super::key::Key;
 use super::{read, Result};
 use crate::kv;
-use crate::kv::Store;
 
 #[allow(dead_code)]
 pub struct Write<'a> {
@@ -46,11 +45,11 @@ impl<'a> Write<'_> {
             .await?)
     }
 
-    pub async fn commit(&mut self) -> Result<()> {
+    pub async fn commit(self) -> Result<()> {
         Ok(self.kvw.commit().await?)
     }
 
-    pub async fn rollback(&mut self) -> Result<()> {
+    pub async fn rollback(self) -> Result<()> {
         Ok(self.kvw.rollback().await?)
     }
 }
@@ -59,6 +58,7 @@ impl<'a> Write<'_> {
 mod tests {
     use super::*;
     use crate::kv::memstore::MemStore;
+    use crate::kv::Store;
 
     #[async_std::test]
     async fn put_chunk() {
@@ -129,18 +129,11 @@ mod tests {
                 let kvr = kv.read().await.unwrap();
                 assert!(!kvr.has("c/h1/d").await.unwrap());
 
-                // After finalize, nothing should work because the tx is closed.
-                // TODO: Better to test for precise error.
                 if commit {
                     w.commit().await.unwrap();
                 } else {
                     w.rollback().await.unwrap();
                 }
-                assert!(w.has_chunk("").await.is_err());
-                assert!(w.get_chunk("").await.is_err());
-                assert!(w.put_chunk(&c).await.is_err());
-                assert!(w.get_head("").await.is_err());
-                assert!(w.set_head("", "").await.is_err());
             }
 
             // The data should now be visible if it was committed.
--- a/src/kv/idbstore.rs
+++ b/src/kv/idbstore.rs
@@ -1,10 +1,20 @@
 use crate::kv::{Read, Result, Store, StoreError, Write};
+use async_std::sync::{Arc, Condvar, Mutex};
 use async_trait::async_trait;
 use futures::channel::oneshot;
+use futures::future::join_all;
 use log::warn;
+use std::collections::HashMap;
 use wasm_bindgen::closure::Closure;
 use wasm_bindgen::{JsCast, JsValue};
-use web_sys::IdbDatabase;
+use wasm_bindgen_futures::spawn_local;
+use web_sys::{IdbDatabase, IdbObjectStore, IdbTransaction};
+
+impl From<String> for StoreError {
+    fn from(err: String) -> StoreError {
+        StoreError::Str(err)
+    }
+}
 
 impl From<JsValue> for StoreError {
     fn from(err: JsValue) -> StoreError {
@@ -70,11 +80,11 @@ impl IdbStore {
 #[async_trait(?Send)]
 impl Store for IdbStore {
     async fn read<'a>(&'a self) -> Result<Box<dyn Read + 'a>> {
-        return Err(StoreError::Str("Not supported".to_string()));
+        Ok(Box::new(ReadTransaction::new(self)?))
     }
 
     async fn write<'a>(&'a self) -> Result<Box<dyn Write + 'a>> {
-        return Err(StoreError::Str("Not supported".into()));
+        Ok(Box::new(WriteTransaction::new(self)?))
     }
 
     async fn put(&mut self, key: &str, value: &[u8]) -> Result<()> {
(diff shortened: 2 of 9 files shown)


aboodman:
> o This also switches Store::commit()/rollback() to move the Store in,
letting the compiler prevent any subsequent uses. The same pattern had
to be applied to dag as well otherwise it looked like a dag::Write
still held a reference. Removed some tests that are now meaningless
and the compiler wouldn't let stand anymore

Duh. didn't even think about the fact I could use the same technique I suggested to avoid those tests. Think I was pretty bleary when I got to unit testing :-/.

aboodman:
> I don't even know what that means, but it compiles with no warnings.

Rust (tm)

src/kv/memstore.rs
```diff
@@ -73,37 +73,26 @@ impl Read for ReadTransaction<'_> {
...
-        self.check_open()?;
         match self.pending.lock().await.contains_key(key) {
             true => Ok(true),
             false => self.rt.has(key).await,
         }
     }
     async fn get(&self, key: &str) -> Result<Option<Vec<u8>>> {
-        self.check_open()?;
         match self.pending.lock().await.get(key) {
             Some(v) => Ok(Some(v.to_vec())),
```
aboodman, line 97: In the `Some` case it would definitely be possible to return a reference to the underlying vector.

In the `None` case, we end up copying the data out of the JS heap into our own vector.

In summary, maybe a TODO to eliminate the copy in both cases?

src/kv/memstore.rs
```diff
@@ -117,16 +106,14 @@ impl Write for WriteTransaction<'_> {
...
         self.pending
             .lock()
             .await
             .insert(key.to_string(), value.to_vec());
         Ok(())
     }
 
-    async fn commit(&mut self) -> Result<()> {
-        self.check_open()?;
+    async fn commit(self: Box<Self>) -> Result<()> {
```
aboodman, line 116: did not even know you could do `self: Box<Self>`. For education, why is it important to make this method only callable on the boxed version? Like what would be wrong with putting it on &Self and then calling it through the Boxed instance.

tests/idbstore.rs
```diff
@@ -0,0 +1,111 @@
...
+    use replicache_client::wasm;
+    use wasm_bindgen_test::wasm_bindgen_test_configure;
+    use wasm_bindgen_test::*;
+
+    wasm_bindgen_test_configure!(run_in_browser);
+
+    async fn new_store() -> Box<dyn Store> {
+        let mut rng = rand::thread_rng();
+        let name = std::iter::repeat(())
+            .map(|()| rng.sample(rand::distributions::Alphanumeric))
```
aboodman, line 14: I think you can also say `map(|_| ...` here? One char shorter!

src/dag/write.rs
```diff
@@ -46,11 +45,11 @@ impl<'a> Write<'_> {
             .await?)
     }
 
-    pub async fn commit(&mut self) -> Result<()> {
+    pub async fn commit(self) -> Result<()> {
```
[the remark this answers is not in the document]
aboodman: In my head I'm going to translate `&mut` to `mutator_lock.acquire()` or something.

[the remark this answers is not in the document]
whiten: Technically I think `self` is mutable here, because it's been moved in, so it's neither a mutable nor a non-mutable reference anymore!

aboodman:
Oh, good point.

whiten:
Need to run for now, but will follow up on this feedback soon. Thank you!

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
+        tx.set_onerror(Some(callback.as_ref().unchecked_ref()));
+        wt.callbacks.push(callback);
+
+        Ok(wt)
+    }
+
+    fn tx_callback(&self, new_state: WriteState) -> Closure<dyn FnMut()> {
+        let pair = self.pair.clone();
+        Closure::once(move || {
+            spawn_local(async move {
```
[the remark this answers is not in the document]
whiten: Good question! We actually don't need it to happen asynchronously, but we do need to get into an async context to await the mutex. I believe the mutex and condvar must be async because we later wish to block on them in an async context, even though we're in a sync context here in the callback.

I changed spawn_local() to instead be task::block_on(), to indicate those no asynchrony is needed.

I'm not 100% sure why Arc is required, but it's present in all of the CondVar examples, sync and non:
https://doc.rust-lang.org/beta/std/sync/struct.Condvar.html, https://docs.rs/async-std/1.6.2/async_std/sync/struct.Condvar.html. I think it's there to support the `.clone()` operation, which the compiler complains about if it's not present, probably because of the `move`s here. I'm not sure if this is protecting the Mutex or CondVar themselves, or just the anonymous struct that holds the two of them, or what, but I stuck with the examples.

src/kv/memstore.rs
whiten: This is true for MemStore, but not IdbStore, and these function signatures are defined by the traits. Given that, I don't think there's a good change to make here, even though you're right that the memory seems like it could be owned by the MemStore.

Even if we didn't have a trait though, wouldn't it be weird if you did a get(), then a put() over the same key, and the variable holding the get result had changed (or disappeared)? I don't see how you could define the memory to be owned by the store if it were to outlive the next write to the store without acrobatics.

src/kv/memstore.rs
whiten: I don't know if this is right or not, but here's what I know: if this signature is `self: Self`, then we get `error[E0161]: cannot move a value of type dyn kv::Write: the size of dyn kv::Write cannot be statically determined` at the callers, because it's a trait. If we defined this on a reference instead it would work, but we wouldn't get the "the call ends the lifetime" behavior that we do get with this, and which seems nice for the API.

Any further thoughts?

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
 }
+
+struct ReadTransaction {
+    tx: IdbTransaction,
+    store: IdbObjectStore,
+}
+
+impl ReadTransaction {
+    fn new(idb: &IdbStore) -> Result<ReadTransaction> {
+        let tx = idb.idb.transaction_with_str(OBJECT_STORE)?;
```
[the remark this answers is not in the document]
whiten: Totally reasonable. Changed to `store.db.`...

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
+            tx: tx,
+        })
+    }
+}
+
+#[async_trait(?Send)]
+impl Read for ReadTransaction {
+    async fn has(&self, key: &str) -> Result<bool> {
+        let request = self.store.count_with_key(&key.into())?;
+        let (sender, receiver) = oneshot::channel::<()>();
```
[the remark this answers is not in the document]
whiten: Good suggestion, did it. I wasn't able to get farther than you'll see because of ownership stuff with `callback`.

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
+        let callback = Closure::once(move || {
+            if let Err(_) = sender.send(()) {
+                warn!("oneshot send failed");
+            }
+        });
+        request.set_onsuccess(Some(callback.as_ref().unchecked_ref()));
+        request.set_onerror(Some(callback.as_ref().unchecked_ref()));
+        receiver.await?;
+        Ok(match request.result()?.as_f64() {
```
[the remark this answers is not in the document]
whiten: 👍

whiten:
Changes for feedback in https://github.com/rocicorp/repc/pull/29.

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
+            if let Err(_) = sender.send(()) {
+                warn!("oneshot send failed");
+            }
+        });
+        request.set_onsuccess(Some(callback.as_ref().unchecked_ref()));
+        request.set_onerror(Some(callback.as_ref().unchecked_ref()));
+        receiver.await?;
+        Ok(match request.result()? {
+            v if v.is_undefined() => None,
+            v => Some(js_sys::Uint8Array::new(&v).to_vec()),
```
arv, line 206: Not yet clear about the details here but I'm concerned about extra copies of the data here.

arv: Especially if go from idb to to rust and then back to js, I wonder if we can pass along opaque values? Probably not at this layer but worth thinking about in the end

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
+    Open,
+    Committed,
+    Aborted,
+    Errored,
+}
+
+struct WriteTransaction {
+    rt: ReadTransaction,
+    pending: Mutex<HashMap<String, Vec<u8>>>,
+    pair: Arc<(Mutex<WriteState>, Condvar)>,
```
arv, line 222: Is there a more descriptive name?

src/kv/idbstore.rs
```diff
@@ -145,3 +155,214 @@ impl Store for IdbStore {
...
+        match *lock.lock().await {
+            WriteState::Committed | WriteState::Aborted => return Ok(()),
+            _ => (),
+        }
+
+        self.rt.tx.abort()?;
+        let state = cv
+            .wait_until(lock.lock().await, |state| *state != WriteState::Open)
+            .await;
+        match self.rt.tx.error() {
```
arv, line 359: Can't these be written using `if let Some(e) = self.rt.tx.error() {`

aboodman: yep

whiten: Clippy for the win!
https://github.com/rocicorp/repc/commit/b67ed4c80066c8aef66550e38f44df8e8b21e74f#diff-e83d091a2db9fc52230899a70d97e3a4R299-R300

arv: Sweet. At least I picked up some best practice along my way!

PR #32, opened 2020-07-08
Introduce Clippy tool to help keep our Rust clean.

whiten:
We should talk as a group about whether we want to embrace this tool,
but I find things like this useful. I think I'd favor embracing it, and
then globally disabling warnings we find unhelpful so we keep the code
in a clean state (per our config) and can always rely on this to find
simplification or real issues like with the bit shift noted below.

o Uses IdbStore::oneshot_callback() in one more place in idbstore.rs.

o Introduces tool/flatc.sh to document how our Flatbuffers are built.
  Also prepends produced file to silence a Clippy warning in the
  generated Flatbuffer code.

o All other changes here are in response to warnings from Clippy. A few
  notable ones:

  o Bug: Parenthesize bit shift operations--+/- are higher precedence in Rust
    (and C), leading our patterns to be like 0x4000 rather than 0x3fff.
  o Remove various unneeded lifetime annotations.
  o Switch a few &String to &str, which it said removed an allocation.
  o Always prefer { foo } over { foo: foo }.
  o Remove all end-of-function return statements.
  o Remove Hash.to_string(), which is materialized implicitly if the
    fmt::Display trait is implemented. Just implement that instead.
  o Change 'match' statements that only catch one condition to 'if' with
    capture.


src/prolly/chunker.rs
```diff
@@ -46,10 +46,10 @@ mod tests {
 
     #[test]
     fn test_hash_byte() {
-        const S: &str = "NaN";
```
whiten, line 49: @aboodman, I wanted to get a sanity check on the changes here and above, which make the mask pattern into the N bits being set where N is the shift amount, versus just having the N + 1th bit set and no others.

Do we have any golden test data from elsewhere that we can use for this test?

PR #38, opened 2020-07-12
Add schema for db package!

aboodman:
🎉

I just wanted to write this out. Feels good to be back up to the top level.


aboodman:
@phritz I used the nomenclature from your most recent design doc changes. I still don't have a better name for this layer than db, suggestions welcome.

whiten:
Just wanted to point out `tool/flatc.sh`, which you should wire this into. If you want to just copy and paste for now I can come back and turn it into a loop over the set of found `.fbs` source later.

It's interesting (but I guess makes sense) that neither `cargo fmt` nor `cargo clippy` notice this new module because it isn't referenced by `src/lib.rs`.

src/db/commit.fbs
```diff
@@ -0,0 +1,32 @@
+namespace commit;
+
+table LocalMeta {
```
phritz, line 3: can we put comments in the definition files? if so I think it would be helpful. f

aboodman:
Comments addressed in https://github.com/rocicorp/repc/pull/38/commits/8a3713adb1b37ce486ee518c06a2577b40babbde

PR #40, opened 2020-07-13
smol is dead code and

arv:


arv:
I assume the idea was to use this in a test but it is not used in any test

arv:
@whiten Where does the **Rust / Wasm Size** go?

aboodman:
Oh yeah I guess I was thinking I'd use it in a test and then didn't. Thanks.

whiten:
> @whiten Where does the **Rust / Wasm Size** go?

@arv, look for the "Artifacts" link at https://github.com/rocicorp/repc/pull/40/checks. (I wish I knew how to put it somewhere better.)

Issue #35, opened 2020-07-09
Add unit testing for dag::Read


aboodman:
We have extensive unit testing for Write transactions but *cough* somebody didn't do the same for Read.

PR #41, opened 2020-07-13
add test for Dag::Read::get_chunk

phritz:
progress towards #35

### DIFF
--- a/src/dag/read.rs
+++ b/src/dag/read.rs
@@ -63,30 +63,76 @@ mod tests {
 
     #[async_std::test]
     async fn test_has_chunk() {
-        async fn test(hash: &str, del_haskey: bool, expect_has: bool) {
+        async fn test(hash: &str, del_key: bool, expect_has: bool) {
+            let k = "present";
             let kv = MemStore::new();
             let kvw = kv.write().await.unwrap();
             let kvr = kv.read().await.unwrap();
             let r = Read { kvr };
 
-            kvw.put(&Key::ChunkData("haskey").to_string(), &vec![0u8, 1])
+            kvw.put(&Key::ChunkData(k).to_string(), &vec![0u8, 1])
                 .await
                 .unwrap();
             kvw.commit().await.unwrap();
             assert_eq!(expect_has, r.has_chunk(&hash).await.unwrap());
 
-            // Mayve test isolation: other txs should not affect what the read tx, which is
+            // Maybe test isolation: other txs should not affect what the read tx, which is
             // still open, sees.
-            if del_haskey {
+            if del_key {
                 let kvw2 = kv.write().await.unwrap();
-                kvw2.del("haskey".into()).await.unwrap();
+                kvw2.del(k).await.unwrap();
                 kvw2.commit().await.unwrap();
                 assert_eq!(expect_has, r.has_chunk(&hash).await.unwrap());
             }
         }
 
-        test("haskey", false, true).await;
-        test("haskey", true, true).await;
-        test("missing", false, false).await;
+        test("present", false, true).await;
+        test("present", true, true).await;
+        test("no such hash", false, false).await;
+    }
+
+    #[async_std::test]
+    async fn test_get_chunk() {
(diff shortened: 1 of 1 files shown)


src/dag/read.rs
```diff
@@ -63,30 +63,76 @@ mod tests {
...
+        test("present", false, true).await;
+        test("present", true, true).await;
+        test("no such hash", false, false).await;
+    }
+
+    #[async_std::test]
+    async fn test_get_chunk() {
+        async fn test(data: Vec<u8>, refs: &[&str], get_same_chunk: bool) {
+            let kv = MemStore::new();
+            let cloned_data = data.clone();
```
aboodman, line 98: [cbb] Eliminate this local by calling `clone` inline at 100.

src/dag/read.rs
```diff
@@ -63,30 +63,76 @@ mod tests {
...
+                        refs,
+                        got_chunk.refs().unwrap().collect::<Vec<&str>>().as_slice()
+                    );
+                }
+            } else {
+                let got_option = r.get_chunk("no such hash").await.unwrap();
+                assert_eq!(None, got_option);
+            }
+        }
+        test(vec![1u8], &vec!["r1", "r2"], true).await;
```
aboodman, line 134: You can say `vec![1]` (the compiler knows it must be u8 due to parameter type of `test`).

src/dag/read.rs
```diff
@@ -63,30 +63,76 @@ mod tests {
...
+                .await
+                .unwrap();
+            if let Some(meta) = chunk.meta() {
+                kvw.put(&Key::ChunkMeta(chunk.hash()).to_string(), meta)
+                    .await
+                    .unwrap();
+            }
+            kvw.commit().await.unwrap();
+
+            // TODO(phritz) start a write tx here that completes before r.get_chunk
```
aboodman, line 111: 🙏

src/dag/read.rs
```diff
@@ -63,30 +63,76 @@ mod tests {
...
+            // below and ensure it does not affect the result.
+            let kvr = kv.read().await.unwrap();
+            let r = Read { kvr };
+            // TODO(phritz) start a write tx here that completes before r.get_chunk
+            // and ensure it does not affect the result.
+            if get_same_chunk {
```
aboodman, line 117: [minifix]

Chunk implements `PartialEq` so I believe that we can reduce this entire if/else block to something like:

```
let expected: Option<Chunk>::None;
if get_same_chunk {
  expected = Some(chunk);
}

assert_eq!(chunk, r.get_chunk(chunk_hash).await.unwrap());
```

src/dag/read.rs
phritz: well, actually... see incoming PR

src/dag/read.rs
phritz: i tried for 15m to take your suggestion but i can't figure out how to make the compiler happy about getting chunk.hash() out of the chunk so we can use it in get_chunk. if you can make this work lemme know i would like to see solution. (also i can post my attempt if you want.)

aboodman: A way to write it that is close to what I suggested it:

```
            let mut expected = Option::<Chunk>::None;
            let chunk_hash: &str;
            if get_same_chunk {
                expected = Some(chunk);
                chunk_hash = expected.as_ref().unwrap().hash();
            } else {
                chunk_hash = "no such hash";
            }
            assert_eq!(expected, r.get_chunk(chunk_hash).await.unwrap());
```

Here is a perhaps simpler way:

```
            if get_same_chunk {
                let expected = Some(chunk);
                assert_eq!(
                    expected,
                    r.get_chunk(expected.as_ref().unwrap().hash())
                        .await
                        .unwrap()
                );
            } else {
                assert_eq!(None, r.get_chunk("No such chunk").await.unwrap());
            }

```

aboodman: I'm sure there's smaller versions (maybe by making `expected` contain a ref), but running out the door :).

phritz: thanks, went with your first one which is easiest for me to read. i was missing the as_ref() to make it work!

PR #55, opened 2020-07-17
use a rwlock to protect the kv store

### LINKED ISSUE #42: memstore txs are not properly isolated
https://rocicorp.slack.com/archives/CMQQ9EDML/p1594671475124600

phritz:
as of this pr writes are mutually exclusive with reads and at most one write tx can be outstanding at a time.

fixes #42 and contributes to #35


src/dag/read.rs
```diff
@@ -135,8 +121,8 @@ mod tests {
...
         }
-        test(vec![1u8], &vec!["r1", "r2"], true).await;
-        test(vec![1u8], &vec![], true).await;
-        test(vec![1u8], &vec!["r1", "r2"], false).await;
+        test(vec![1], &vec!["r1", "r2"], true).await;
+        test(vec![1], &vec![], true).await;
+        test(vec![1], &vec!["r1", "r2"], false).await;
     }
-}
+}
```
aboodman, line 128: [nit] missing final newline.

src/kv/memstore.rs
```diff
@@ -24,47 +24,49 @@ impl Default for MemStore {
...
     async fn write<'a>(&'a self) -> Result<Box<dyn Write + 'a>> {
-        Ok(Box::new(WriteTransaction::new(self)))
+        let guard = self.map.write().await;
+        Ok(Box::new(WriteTransaction::new(guard)))
     }
 }
 
 struct ReadTransaction<'a> {
-    store: &'a MemStore,
+    guard: RwLockReadGuard<'a, HashMap<String, Vec<u8>>>,
```
aboodman, line 38: `guard` seems like kind of an odd name for this since it actually contains the data too. Maybe I'd just call it `map`, or `guarded_map` ?

src/kv/memstore.rs
```diff
@@ -24,47 +24,49 @@ impl Default for MemStore {
...
+    guard: RwLockWriteGuard<'a, HashMap<String, Vec<u8>>>,
     pending: Mutex<HashMap<String, Option<Vec<u8>>>>,
 }
 
 impl WriteTransaction<'_> {
-    fn new(store: &MemStore) -> WriteTransaction {
+    fn new(guard: RwLockWriteGuard<'_, HashMap<String, Vec<u8>>>) -> WriteTransaction {
         WriteTransaction {
-            rt: ReadTransaction { store },
+            guard: guard,
```
aboodman, line 69: nit: you can just say `guard`. Also I think `cargo clippy` will whine about this.

src/kv/memstore.rs
```diff
@@ -76,15 +78,18 @@ impl Read for WriteTransaction<'_> {
...
+            None => Ok(self.guard.contains_key(key)),
         }
     }
 
     async fn get(&self, key: &str) -> Result<Option<Vec<u8>>> {
         match self.pending.lock().await.get(key) {
             Some(Some(v)) => Ok(Some(v.to_vec())),
             Some(None) => Ok(None),
-            None => self.rt.get(key).await,
+            None => match self.guard.get(key) {
```
aboodman, line 89: [taste]: A shorter way to say this final branch is:

```
Ok(self.guard.get(key).map(|v| v.to_vec()))
```

src/kv/memstore.rs
```diff
@@ -108,76 +113,147 @@ impl Write for WriteTransaction<'_> {
...
 
         Ok(())
     }
 
     #[async_std::test]
-    async fn delete() -> std::result::Result<(), StoreError> {
+    async fn test_write_transaction() -> std::result::Result<(), StoreError> {
         let mut ms = MemStore::new();
-        ms.put("bar", "foo".as_bytes()).await?;
+        ms.put("k1", "v1".as_bytes()).await?;
```
aboodman, line 182: [minifix] `b"v1"` instead of `"v1".to_bytes()` -> clippy will tell you this too.

src/kv/memstore.rs
```diff
@@ -108,76 +113,147 @@ impl Write for WriteTransaction<'_> {
...
+        // Test put then commit.
         let wt = ms.write().await?;
-        assert_eq!(true, wt.has("bar").await?);
-        wt.del("bar").await?;
-        assert_eq!(false, wt.has("bar").await?);
-        wt.put("bar", b"overwrite").await?;
-        assert_eq!(true, wt.has("bar").await?);
-        assert_eq!(Some(b"overwrite".to_vec()), wt.get("bar").await?);
-        wt.del("bar").await?;
+        assert_eq!(true, wt.has("k1").await?);
```
aboodman, line 187: `assert!(wt.has(...).await);`

aboodman: Multiple places below.

src/kv/memstore.rs
```diff
@@ -108,76 +113,147 @@ impl Write for WriteTransaction<'_> {
...
+        let rt = wt.as_read();
+        assert_eq!(true, rt.has("k2").await?);
+        assert_eq!(Some(b"new value".to_vec()), rt.get("k2").await?);
 
         Ok(())
     }
+
+    #[async_std::test]
+    async fn test_isolation() {
+        let ms = MemStore::new();
```
aboodman, line 235: [minifix]

`use std::time::Duration` to avoid the repetitions below. Can be local to this function.

aboodman:
Nice. You can use `cargo clippy` locally to find all its suggestions. They are usually quite helpful.

src/kv/memstore.rs
phritz: the long qualifiers dont bother me, probably bc i'm still new and like to see where things come from?

phritz:
thanks, think i addressed em all. we should see if we can get clippy to run in github to annotate prs...

phritz:
@whiten happy to have your review post-merge. and thanks for this change, i learned a lot.

PR #57, opened 2020-07-21
Add benchmarks for db-level dispatch calls.

whiten:
These are novel in that the benchmark loop executes in JS space and each
iteration calls into Rust and transfers the response back to JS.

Example run:

```
% ./tool/benchmark dispatch
Hardware: 8-Core Intel Core i9 2.4 GHz, 32 GB RAM, 1 TB SSD

dispatch::has           483   2,141,490 ns/iter
dispatch::get1024       468   2,229,690 ns/iter   0.46 MB/s
dispatch::get4096       459   2,273,355 ns/iter   1.80 MB/s
dispatch::get16384      438   2,463,367 ns/iter   6.65 MB/s
dispatch::put1024       100  23,429,150 ns/iter   0.04 MB/s
dispatch::put4096       100  35,972,949 ns/iter   0.11 MB/s
dispatch::put16384      100  57,210,150 ns/iter   0.29 MB/s
dispatch::random4096  10855     108,412 ns/iter  37.78 MB/s
```

### DIFF
--- a/src/benches/dispatch.rs
+++ b/src/benches/dispatch.rs
@@ -1,0 +1,147 @@
+use crate::benches::{random_bytes, random_string};
+use crate::dispatch::*;
+use crate::kv::idbstore::IdbStore;
+use crate::kv::Store;
+use wasm_bench::*;
+
+async fn opendb() -> String {
+    let dbname = random_string(12);
+    dispatch(dbname.clone(), "open".into(), "".into())
+        .await
+        .unwrap();
+    dbname
+}
+
+async fn eval(code: &str) {
+    wasm_bindgen_futures::JsFuture::from(js_sys::Promise::resolve(&js_sys::eval(code).unwrap()))
+        .await
+        .unwrap();
+}
+
+#[wasm_bench]
+async fn has(b: &mut Bench) {
+    let dbname = opendb().await;
+
+    let store = IdbStore::new(&dbname).await.unwrap().unwrap();
+    let wt = store.write().await.unwrap();
+    for i in 0..b.iterations() {
+        if i % 2 == 0 {
+            wt.put(&format!("{}", i), &random_bytes(512)).await.unwrap();
+        }
+    }
+    wt.commit().await.unwrap();
+
+    b.reset_timer();
+    eval(&format!(
+        "
+        (async _ => {{
+            for (let i = 0; i < {}; i++) {{
+                await dispatch(\"{}\", \"has\", '{{\"key\": \"' + i + '\"}}');
+            }}
+        }})()",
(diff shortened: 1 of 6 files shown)


src/benches/dispatch.rs
```diff
@@ -0,0 +1,147 @@
...
+    let wt = store.write().await.unwrap();
+    for i in 0..b.iterations() {
+        if i % 2 == 0 {
+            wt.put(&format!("{}", i), &random_bytes(512)).await.unwrap();
+        }
+    }
+    wt.commit().await.unwrap();
+
+    b.reset_timer();
+    eval(&format!(
```
aboodman, line 35: Little concerned about the use of `eval` (assuming that's what's happening under the covers here) deoptimizing.

whiten: Very interested to hear more. What do you mean by that? And yes, this is using JS's eval to construct a benchmark of how long these functions take from a JS context. I should create a parallel copy of some of these that executes solely in wasm to isolate the difference that comes from the JS->wasm hop, but I'll wait for more feedback first.

aboodman: There are a lot of things js can do to accidentally trigger the jit to stop optimizing. i think eval might be one of those things. I don't know a lot about it, just to be wary of it. I'm not sure if it would just de-optimize this loop, or potentially what it's calling into like the marshaling code. @arv might know more.

aboodman: Basically I don't know if using eval() is a good simulation of somebody using the js library, but I know enough to know it's something to look at.

I guess you could find out the answer by running this in a browser and looking in the web inspector at the profiler. I know it marks functions that have been depotimized:

https://github.com/GoogleChrome/devtools-docs/issues/53

arv: This `eval` looks fine since it is only used once at the top level of the computation.

arv: But if there is any kind of iteration of `has` then this is not going to be a realistic benchmark.

src/benches/dispatch.rs
```diff
@@ -0,0 +1,147 @@
...
+    let dbname = opendb().await;
+
+    b.bytes = size;
+    b.reset_timer();
+    eval(&format!(
+        "
+        (async _ => {{
+            var array = new Uint8Array({});
+            for (let i = 0; i < {}; i++) {{
+                window.crypto.getRandomValues(array);
```
arv, line 118: This should not be inside the benchmark. It is probably slow.

arv: And deopts

PR #58, opened 2020-07-21
WIP: test idbstore against Store trait tests

phritz:


src/kv/memstore.rs
```diff
@@ -128,137 +128,24 @@ impl Write for WriteTransaction<'_> {
...
-        assert_eq!(Some(b"final".to_vec()), ms.get("k2").await?);
-
-        // Test as_read.
-        let wt = ms.write().await?;
-        wt.put("k2", b"new value").await?;
-        let rt = wt.as_read();
-        assert!(rt.has("k2").await?);
-        assert_eq!(Some(b"new value".to_vec()), rt.get("k2").await?);
+    async fn test_memstore() -> std::result::Result<(), StoreError> {
+        let ms = &mut MemStore::new();
```
aboodman, line 140: Feels like it will be easy to accidentally forget to add to this list. Can it be mechanised more?

phritz: yes indeed. lets sort out what we're doing with this PR then i will address.

phritz:
good lord, finally there. i didnt see much value in reworking the idbstore constructor to be able to replace the box dyn store with a static version, but could pursue that if you feel it is important.

i feel like someone trying to make the leap from BASIC to C.

PR #59, opened 2020-07-21
Add init_client_id function

### LINKED ISSUE #45: Implement clientID
Various bits of the Replicache sync protocol rely on having a per-client unique ID that is generated internally. See https://github.com/rocicorp/replicache-client/blob/master/db/config.go.

arv:
It currently gets called in open and the client id persists in the
Store.

We store the UUID as a string formatted as
`"0c317bbb-85f9-4854-b47c-18c0660cf6b1"`

Fixes #45


src/dispatch.rs
```diff
@@ -100,17 +104,26 @@ impl Dispatcher {
...
             }
-        }
+        };
+
+        self.connections.insert(
+            req.db_name.clone(),
+            Box::new(dag::Store::new(Box::new(idb_store))),
+        );
+
+        // TODO(arv): If we get None in IdbStore::new that should be an error.
```
whiten, line 126: I think you handled this, right?

src/dispatch.rs
```diff
@@ -100,17 +104,26 @@ impl Dispatcher {
...
-            Ok(v) => {
-                if let Some(kv) = v {
-                    self.connections
-                        .insert(req.db_name.clone(), Box::new(dag::Store::new(Box::new(kv))));
-                }
+        let idb_store = match IdbStore::new(&req.db_name[..]).await {
+            Err(e) => return Err(format!("Failed to open \"{}\": {}", req.db_name, e)),
+            Ok(None) => return Err(format!("Failed to open \"{}\"", req.db_name)),
+            Ok(Some(idb_store)) => {
+                let _client_id = match init_client_id(&idb_store).await {
```
whiten, line 111: I think this should be like:

if let Err(e) = init_client_id(&idb_store).await {
    return Err(...);
}

whiten: (Since you don't care about `_client_id`.)

src/dispatch.rs
```diff
@@ -239,3 +252,35 @@ enum PutError {
     OpenTransactionError(OpenTransactionError),
     CommitError(db::CommitError),
 }
+
+pub async fn init_client_id(s: &dyn Store) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
```
whiten: We don't want this to be public in general, right, it's only for testing?

I wish I knew better what the cost of expanding the interface is, but I feel uncomfortable with it at first blush. Did you try writing the tests below without putting them in a `mod test` so they were black box?

src/dispatch.rs
arv: I want to change IdbStore::new to return an Error but I'm OK dealing with it here.

src/dispatch.rs
```diff
@@ -92,6 +94,8 @@ struct Dispatcher {
     connections: HashMap<String, Box<dag::Store>>,
 }
 
+const CID_KEY: &str = "cfg/cid";
```
aboodman, line 97: Let's use `sys/` -- more general.

src/dispatch.rs
arv: Good point

src/dispatch.rs
arv: We will care about client_id in the future

arv:
Before:

```
replicache_client.js: 18051
replicache_client.js.br: 3775
replicache_client_bg.wasm: 214316
replicache_client_bg.wasm.br: 73486
```

After:

```
replicache_client.js: 19122
replicache_client.js.br: 3909 (+3.5%)
replicache_client_bg.wasm: 228957
replicache_client_bg.wasm.br: 78373 (+6.6%)
```

src/dispatch.rs
```diff
@@ -239,3 +252,35 @@ enum PutError {
     OpenTransactionError(OpenTransactionError),
     CommitError(db::CommitError),
 }
+
+async fn init_client_id(s: &dyn kv::Store) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
```
aboodman, line 256: The pattern I've been playing with for errors is to have a union that describes the potential errors for basically each function. Then if there is an internal error, I wrap the union value around that internal error (manually, there is no From implementation or boilerplate).

The result is basically a tree of strongly-typed errors that compactly describe every error that can happen in the entire call graph, which is pretty neat!

So for this, I'd say:

```
enum InitClientIDError {
    GetClientIDErr(kv::Error),
    OpenWriteTxErr(kv::Error),
    PutClientIDErr(kv::Error),
    CommitTxErr(kv::Error),
}
```

at the call sites, there's a pretty clean pattern:

```
let cid = s.get(CID_KEY).await.map_err(InitClientIDError::GetClientIDErr)?;
```

... or you can `use InitClientIDError::*` to make it shorter:

```
let cid = s.get(CID_KEY).await.map_err(GetClientIDErr)?;
```

I guess trying to extract some of the things I like about this:

- Every distinct error that can happen is strongly typed and visible to callers. You don't have to decide ahead of time which errors are important enough to give distinct values to make "catchable", and there's no separate machinery for that. And you can clearly see at each call site "what are the reasons this can fail?".
- All errors, no matter how obscure, get a relatively descriptive message by default
- You can easily map all errors to their source *and call graph* without line numbers
- Errors values are compact at runtime (just some integers) and stack allocated (I guess this doesn't matter though if errors are rare)

The downside is it that it's a bit burdensome to describe these enums. It would be nice to have a macro or some other machinery that generated them, but so far it's not that big a deal.

An alternative that would be less machinery would be the Go way, basically of prepending the context error message to the received error message and propagating it. But that loses the "what are the reasons this can fail?" transparency and also:

* There's no guarantee of string uniqueness so I always worry about tracking down the source of these errors
* There's mental overhead to wondering whether to wrap errors. If you don't wrap enough then it can be hard to figure out where an error came from. If you wrap too much the error messages get super long and hard to read.

Anyway I'm curious what you all think of adopting this pattern. So far I have enjoyed it. Try triggering an error in some of the code that uses it and looking at the resulting errors.

Anyway not a blocker for this, but we should discuss and narrow in on an error handling strategy that we use consistently throughout.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ sha2 = "0.8.1"
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4.13"
 wee_alloc = "0.4.5"
+uuid = { version = "0.8", features = ["v4", "wasm-bindgen"] }
```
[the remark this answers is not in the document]
arv: Found this... https://github.com/tmtmtoo/uuid-v4-wasm

Looks very simple but maybe guys enough

arv:
Converting this back to a draft...

Main new thing is that we now use our "own" UUIDV4 using window.crypto.getRandom... Based on https://github.com/tmtmtoo/uuid-v4-wasm but no need to add the extra level of XOR shift PRG on top of window.crypto.

I don't know if I will be able to finish this PR

src/dispatch.rs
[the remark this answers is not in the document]
aboodman: Ah, my mistake. Thanks.

src/uuid/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+fn make_random_numbers(mut numbers: &mut [u8]) {
+    web_sys::window()
+        .expect("window is not available")
+        .crypto()
+        .expect("window.crypto is not available")
+        .get_random_values_with_u8_array(&mut numbers)
+        .expect("window.crypto.getRandomValues not available");
+}
+
+#[cfg(not(target_arch = "wasm32"))]
```
aboodman, line 22: Don't we need one for not-wasm?

arv:
Before:

```
replicache_client.js: 18051
replicache_client.js.br: 3775
replicache_client_bg.wasm: 214316
replicache_client_bg.wasm.br: 73486
```

After:

```
replicache_client.js: 18829
replicache_client.js.br: 3896 + 3.2%
replicache_client_bg.wasm: 215301
replicache_client_bg.wasm.br: 74509 +1.4%
```

Still seems strange that the size increased that much?

aboodman:
Maybe this is the first usage of random.

aboodman:
Superceded by #169

PR #70, opened 2020-08-03
working TOY queue version

phritz:


src/experimental/queue.rs
```diff
@@ -0,0 +1,372 @@
...
+// database ("store" in the code here) has its own connection_loop.
+// The dispatch_loop sends requests for a given db to its connection_loop.
+//
+// The connection_loop keeps a queue of incoming requests. It keeps at most
+// one write tx open at a time and when a write tx is open no other txs may run.
+// In order not to deadlock the connection_loop scans the queue for requests
+// for the open write tx if any, and runs them ahead of the rest of the queue.
+// 
+// Caveats:
+// - the database just stores a bool
```
aboodman, line 26: ship it

src/experimental/queue.rs
```diff
@@ -0,0 +1,372 @@
...
+// and locks.
+// 
+// Dispatch happens in two layers. dispatch_loop is the first layer, 
+// responsible for opening and managing connections to databases. Each 
+// database ("store" in the code here) has its own connection_loop.
+// The dispatch_loop sends requests for a given db to its connection_loop.
+//
+// The connection_loop keeps a queue of incoming requests. It keeps at most
+// one write tx open at a time and when a write tx is open no other txs may run.
+// In order not to deadlock the connection_loop scans the queue for requests
```
aboodman, line 22: I considered this idea but could not wrap my head around whether it was correct to eagerly process messages for a write transaction ahead of other messages destined for other transactions.

src/experimental/queue.rs
```diff
@@ -0,0 +1,372 @@
...
+    query_ch: Sender<Request>,
+    // Signal to quit from dispatch_loop.
+    quit_ch: Sender<()>,
+}
+
+async fn dispatch_loop(req_ch_rec: Receiver<Request>, quit_ch_rec: Receiver<()>) {
+    let mut conns: HashMap<String, Connection> = HashMap::new();
+
+    loop {
+        match quit_ch_rec.try_recv() {
```
aboodman, line 117: Huh. This is very go-y.

src/experimental/queue.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+// One per connection / database. Handles requests for this db.
+async fn connection_loop(req_ch_rec: Receiver<Request>, quit_ch_rec: Receiver<()>, mut store: Store) {
+    // We buffer Requests to the db so we can allow Requests to the current
+    // write tx (if any) to jump the queue.
+    let mut queue = VecDeque::new();
+
+    // If write_tx is None we are executing reads.
+    let mut write_tx: Option<WriteTx> = None;
```
aboodman, line 207: one meta-question I had as I was working through this:

given the model that we are strictly serialized, are the read transactions doing anything for us?

src/experimental/queue.rs
```diff
@@ -0,0 +1,372 @@
...
+    let mut next_tx_id = 0u64;
+
+    loop {
+        // Check if we have received the quit signal.
+        match quit_ch_rec.try_recv() {
```
aboodman, line 212: I think there is probably a lighter-weight way to shut these things down than the go-style way.

src/experimental/queue.rs
```diff
@@ -0,0 +1,372 @@
...
+                // TODO log an error
+                return;
+            }
+            Ok(_) => return,
+        }
+
+        // We want to wait asynchronously for a new request (and not spin) if
+        // we have nothing to do. We have nothing to do if:
+        //   A) the queue is empty or
+        //   B) we have an open write tx but zero Requests for it in the queue
```
aboodman, line 224: I do not think that this is quite correct. If a read transaction is open while a write transactions starts, that read transaction will stall until the write transaction completes. Worse, the read transaction will read an inconsistent view of the world -- part before the write tx and part after.

I think that we need to do something more like I proposed in the sketch -- we keep track of open read transactions too -- we can either be in a mode where we're dispatching to n open read tx, or a mode where we are dispatching to one open write tx. Before we start the write tx, all open read tx must complete, just as you have us waiting for the write tx now.

src/experimental/queue.rs
aboodman: Ah, i remember. yes we need them: https://github.com/rocicorp/repc/pull/70/files#r464167947

src/experimental/queue.rs
phritz: right. same technique can work for read tx.

aboodman:
As for deadlock: we have been aware of this problem at higher levels of the stack and there was extensive discussion on it: https://github.com/rocicorp/replicache/issues/27#issuecomment-611159927.

I am happy with the mitigations we landed at there, and I'm not sure what else is realistically possible. Alternative ideas welcome.

Issue #44, opened 2020-07-14
Scan


aboodman:
Implement tx.scan -- see https://github.com/rocicorp/replicache-client/blob/master/db/scan.go

aboodman:
If this gets started before #54 then just implement them inside dispatch along side existing get/has/put: https://github.com/rocicorp/repc/blob/master/src/dispatch.rs#L154.

PR #75, opened 2020-08-06
Implement db::scan. Still need to integrate into embed.

aboodman:
Toward #44

### DIFF
--- a/src/db/scan.rs
+++ b/src/db/scan.rs
@@ -1,0 +1,502 @@
+use crate::prolly;
+
+#[allow(dead_code)]
+pub struct ScanKey<'a> {
+    value: &'a [u8],
+    exclusive: bool,
+}
+
+#[allow(dead_code)]
+pub struct ScanBound<'a> {
+    // TODO: Make these two fields exclusive?
+    key: Option<ScanKey<'a>>,
+    index: Option<u64>,
+}
+
+#[allow(dead_code)]
+pub struct ScanOptions<'a> {
+    // TODO: Make these two fields exclusive?
+    prefix: Option<&'a [u8]>,
+    start: Option<ScanBound<'a>>,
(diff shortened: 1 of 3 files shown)


src/db/scan.rs
```diff
@@ -0,0 +1,502 @@
...
+
+#[allow(dead_code)]
+pub struct ScanKey<'a> {
+    value: &'a [u8],
+    exclusive: bool,
+}
+
+#[allow(dead_code)]
+pub struct ScanBound<'a> {
+    // TODO: Make these two fields exclusive?
```
phritz, line 11: I have a preference for only having TODOs for things that we are actually going to do, and soon. Otherwise they just sit there forever. If this doesn't meet the bar maybe just 
```
// These two fields are mutually exclusive.
```
and we're done.

src/db/scan.rs
```diff
@@ -0,0 +1,502 @@
...
+    let key_met = |key: &[u8]| {
+        if exclusive {
+            key > from_key
+        } else {
+            key >= from_key
+        }
+    };
+
+    let mut index = 0;
+    while it.peek().is_some() {
```
phritz, line 60: can peek block, eg going to disk?

src/db/scan.rs
aboodman: Me too. The situation with this code is actually the reverse -- the fields are *not* mutually exclusive currently. I've updated the comment to document the current situation and remove the TODO:

https://github.com/rocicorp/repc/pull/103

src/db/scan.rs
aboodman: Not currently. The entire read API of prolly::Map is synchronous currently because the map is held in memory. Once prolly trees are implemented all the methods will become asynchronous.

PR #76, opened 2020-08-07
Factor out db::Read and teach embed layer about read transactions

aboodman:


aboodman:
@whiten - I know you're taking a little break, don't feel obligated, but reviewing you in case you are curious.

FYI I'm find the complexity around `execute` because of the dual lifetime params hard to keep in my brain. I was going to experiment in subsequent PRs with removing the callback structure and doing something dumber and more c-like.

whiten:
> FYI I'm find the complexity around `execute` because of the dual lifetime params hard to keep in my brain. I was going to experiment in subsequent PRs with removing the callback structure and doing something dumber and more c-like.

Can you expand here a bit? Do you mean putting the deserialize and serialize steps back into each function, or something else? One of my concerns with Rust is that it's so hard to write (and grok) function signatures that it pushes towards repeating code rather than factoring stuff out. That's not a reason to leave stuff that's hard to read in place, but I'm curious if you're having a similar experience. Thanks.

aboodman:
On Tue, Aug 11, 2020 at 8:43 AM Nate White <notifications@github.com> wrote:

> FYI I'm find the complexity around execute because of the dual lifetime
> params hard to keep in my brain. I was going to experiment in subsequent
> PRs with removing the callback structure and doing something dumber and
> more c-like.
>
> Can you expand here a bit? Do you mean putting the deserialize and
> serialize steps back into each function, or something else?
>
Yeah I just wanted to play and see if I could find some other way to factor
this that would basically reduce the number of lifetime params, because I
find that the hardest part of Rust. Some of the things I tried or
considered:

- Can I express the dispatch loop as basically just a simple loop with a
case statement and no callbacks, no abstraction?  Yes, but the fact that
each handler has its own request/response types complicates that, basically
means you'd have to at least duplicate the *call* to serialize and
deserialize. Also the fact that some of the handlers require transactions
and others don't means you're going to have to conditionally create the
transaction inside the loop.

- Can I express the handlers as a trait with associated types for
request/response/error? Yes, but because async is not supported in traits
we have to type the handler to return Future not async. But `impl Future`
is not supported in the return type of a trait....

- OK what if I make the trait generic and have the return future be one of
the types? (Haven't tried this one yet)

- Or what if I just give up and use the async-trait crate (Haven't tried
this one yet)

- Or what if I use some RTTI and store the request/response/error as `any`?

- Macro?

> One of my concerns with Rust is that it's so hard to write (and grok)
> function signatures that it pushes towards repeating code rather than
> factoring stuff out. That's not a reason to leave stuff that's hard to read
> in place, but I'm curious if you're having a similar experience. Thanks.
>
Yeah I know what you mean. I would say that it's especially hard to write
dry code where `async` is involved for several reasons:

- async is not allowed in traits, so you can't easily use traits as a
refactoring tool
- async blocks are not supported
- storing async functions in values is only partially supported
- some of this stuff is maybe better supported on the nightly rust stack,
but i've been reluctant to opt into that (maybe wrongly?)

I'm not sure if the right solution is just to repeat yourself more, to use
more experimental features, use macros, etc. I was kind of using this area
of the code as a place to play and try a few different things on for size.

I don't mean it as a critique (though I can see it did sound that way) --
just like, this is harder than expected, and the easiest way for me to
think through it is by typing.

whiten:
Yeah, all of that makes total sense, and I definitely appreciate the feeling of wanting to just code on some different options to see how they fall out, so feel free to have at it.

In terms of stable vs. experimental stuff, this is definitely a tough area. My feeling is that we've bit the bullet on `async_trait`, and I'm comfortable with that, as it's an obvious missing piece in the language, and the workaround that that crate exposes is unlikely to be too different from what the eventual language feature will offer. Similarly, I've now embraced the `AsyncFnX` macros, and so for these limited things that we've opted in to, I think we should go to town and see what we can make work.

Nightly is tempting, but I think it's a Pandora's Box, as it means opting into the daily bug stream, likely just for one of the ~sixty features that are in nightly and haven't graduated yet. I think it's totally fair to jump over to it to get a feel for what may be coming down the line, I'd just be concerned about making our main branch depend on it. I think I'd feel differently if the feature flags were more like "stable release plus a set of patches" rather than "nightly build with feature X turned on", but that's not what's on offer. I think that last statement may be why I feel more comfortable with things like the `async_trait` crate that try to backport a coming feature, as I expect their scope to be more constrained.

phritz:
I don't have concrete suggestions but  I do have a preference for repetitious code over more advanced language features in rust, at least for the moment. I don't find that repetitious code slows me down very much I do find that lifetime parameters and more advanced language features definitely do slow me down.

PR #78, opened 2020-08-11
Remove a bunch of #[allow(dead_code)] annotations.

whiten:
In a couple places, these were moved to be more specific, e.g.:
o dag::Read::has_chunk()
o db::Read::dag_read
o db::Read::scan()

In kv::IdbStore, I exchanged these dead_code annotations for a leading
understore on the RwLock...Guard variables, as these aren't dead,
they're just unreferenced placeholders to keep a lock alive.

### DIFF
--- a/src/dag/read.rs
+++ b/src/dag/read.rs
@@ -8,7 +8,6 @@ pub struct OwnedRead<'a> {
     kvr: Box<dyn kv::Read + 'a>,
 }
 
-#[allow(dead_code)]
 impl<'a> OwnedRead<'a> {
     pub fn new(kvr: Box<dyn kv::Read + 'a>) -> OwnedRead {
         OwnedRead { kvr }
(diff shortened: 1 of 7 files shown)


aboodman:
How do you find these? is there some tool that tells you? Or is it just a particular thing that bugs you and so you keep an eye out for. In any case, thanks :).

whiten:
> How do you find these? is there some tool that tells you? Or is it just a particular thing that bugs you and so you keep an eye out for. In any case, thanks :).

I don't have a tool, I just remove these and test. I don't know why they stand out to me, but they do, and I just happened to notice some in https://github.com/rocicorp/repc/pull/76 that seemed like they couldn't be necessary, so I did a quick pass.

http://www.mypy-lang.org/, a static analysis tool for Python, does have such a `warn_unused_ignores` flag, and I grew to like knowing that the ignores remaining in the code were meaningful.

Issue #43, opened 2020-07-14
Implement sync!


aboodman:
Once we have implemented #54 we can finally, gloriously, implement the Replicache sync protocol! What a fine day that will be. Actually sync can be started ahead of #54 -- implementing pull doesn't really require transactions at all. And even with push, I suppose you could do most of it against the temporary transactions we are creating for db::put().

To be clear this is basically porting https://github.com/rocicorp/replicache-client/blob/master/db/sync.go to Rust. I thought there was a long issue somewhere documenting how this worked that @phritz wrote, but can't find it now.

aboodman:
Closing this in favor of the myriad finer-grained bugs @phritz opened.

PR #79, opened 2020-08-12
Browser fetch

phritz:
implement browser fetch and the skeleton of beginsync. all it does is attempt to pull.

there is no test for the begin sync or pull code yet. that comes next when i implement http::rust_fetch which means we can write unit tests for pull and sync against an http server.

progress towards #43

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -1,4 +1,5 @@
 use super::dispatch::Request;
+use super::sync;
 use super::types::*;
 use crate::dag;
 use crate::db;
@@ -62,6 +63,7 @@ async fn connection_future<'a, 'b>(
         "openTransaction" => execute(do_open, store, txns, req).await,
         "commitTransaction" => execute(do_commit, store, txns, req).await,
         "closeTransaction" => execute(do_abort, store, txns, req).await,
+        "beginSync" => execute(do_begin_sync, store, txns, req).await,
         "close" => {
             req.response.send(Ok("".into())).await;
             return UnorderedResult::Stop();
@@ -270,6 +272,18 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
     }?;
     write.put(req.key.as_bytes().to_vec(), req.value.into_bytes());
     Ok(PutResponse {})
+}
+
+// This fn just forwards to sync::begin_sync at this point. execute() requires its
+// do_ers to take txns even if they don't need it so this fn also helps begin_sync
+// have a simpler signature.
+async fn do_begin_sync<'a, 'b>(
+    _store: &'a dag::Store,
+    _txns: &'b TxnMap<'a>,
+    req: BeginSyncRequest,
+) -> Result<BeginSyncResponse, sync::BeginSyncError> {
+    let begin_sync_response = sync::begin_sync(&req).await?;
+    Ok(begin_sync_response)
 }
 
 #[derive(Debug)]
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -278,8 +278,8 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
 // do_ers to take txns even if they don't need it so this fn also helps begin_sync
 // have a simpler signature.
 async fn do_begin_sync<'a, 'b>(
-    _store: &'a dag::Store,
-    _txns: &'b TxnMap<'a>,
+    _: &'a dag::Store,
+    _: &'b TxnMap<'a>,
     req: BeginSyncRequest,
 ) -> Result<BeginSyncResponse, sync::BeginSyncError> {
     let begin_sync_response = sync::begin_sync(&req).await?;
--- a/src/embed/http.rs
+++ b/src/embed/http.rs
@@ -26,7 +26,11 @@ pub async fn browser_fetch(
     let web_sys_req = web_sys::Request::new_with_str_and_init(&http_req.uri().to_string(), &opts)?;
     let h = web_sys_req.headers();
     for (k, v) in http_req.headers().iter() {
-        h.set(k.as_ref(), v.to_str().unwrap())?;
+        h.set(
+            k.as_ref(),
+            v.to_str()
+                .map_err(|e| FetchError::new(&format!("{:?}", e)))?,
+        )?;
     }
 
     let window = web_sys::window().ok_or_else(|| FetchError::new("could not get window"))?;
@@ -50,6 +54,7 @@ pub async fn browser_fetch(
 // FetchErrors are returned by both the rust and browser versions of fetch. Since
 // lower level errors in each case will be coming from two different places I implemented
 // FetchError as lossy of the error types underneath: it holds an error string.
+// TODO turn this into an enumerated type.
 #[derive(Debug)]
 pub struct FetchError {
     msg: String,
--- a/src/embed/http.rs
+++ b/src/embed/http.rs
@@ -38,9 +38,9 @@ pub async fn browser_fetch(
     }
     let web_sys_resp: web_sys::Response = js_web_sys_resp.dyn_into().unwrap();
     let body_js_value = JsFuture::from(web_sys_resp.text()?).await?;
-    let resp_body = body_js_value.as_string().ok_or_else(|| FetchError::new(
-        "could not get http response body as string",
-    ))?;
+    let resp_body = body_js_value
+        .as_string()
+        .ok_or_else(|| FetchError::new("could not get http response body as string"))?;
 
     let builder = http::response::Builder::new();
     let http_resp = builder.status(web_sys_resp.status()).body(resp_body)?;
--- a/src/embed/http.rs
+++ b/src/embed/http.rs
@@ -1,0 +1,92 @@
+use js_sys;
+use std::convert::TryFrom;
+use std::error::Error;
+use std::fmt;
+use std::str;
+use wasm_bindgen::JsCast;
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::JsFuture;
+use web_sys::{RequestInit, RequestMode};
+
+// browser_fetch makes and HTTP request over the network via the browser's Fetch API.
+// It is intended to be used in wasm; it won't work in regular rust. The response contains
+// the status and body, it doesn't populate response headers (though it easily could).
+// Since we can't run a web server in wasm land and this fn doesn't work in rust land
+// this fn is tested manually in tests/wasm.rs.
+//
+// TODO timeout/abort
+// TODO log request/response
+pub async fn browser_fetch(
+    http_req: &http::Request<String>,
+) -> Result<http::Response<String>, FetchError> {
+    let mut opts = RequestInit::new();
+    opts.method(http_req.method().as_str());
+    opts.mode(RequestMode::Cors);
+    let js_body = JsValue::from_str(http_req.body());
+    opts.body(Some(&js_body));
+    let web_sys_req = web_sys::Request::new_with_str_and_init(&http_req.uri().to_string(), &opts)?;
+    let h = web_sys_req.headers();
+    for (k, v) in http_req.headers().iter() {
+        h.set(k.as_ref(), v.to_str().unwrap())?;
+    }
+
+    let window = web_sys::window().ok_or(FetchError::new("could not get window"))?;
+    let http_req_promise = window.fetch_with_request(&web_sys_req);
+    let http_req_future = JsFuture::from(http_req_promise);
+    let js_web_sys_resp = http_req_future.await?;
+    if !js_web_sys_resp.is_instance_of::<web_sys::Response>() {
+        return Err(FetchError::new("result from future not a Response"));
+    }
+    let web_sys_resp: web_sys::Response = js_web_sys_resp.dyn_into().unwrap();
+    let body_js_value = JsFuture::from(web_sys_resp.text()?).await?;
+    let resp_body = body_js_value.as_string().ok_or(FetchError::new(
+        "could not get http response body as string",
+    ))?;
+
+    let builder = http::response::Builder::new();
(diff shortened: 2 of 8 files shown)


src/embed/http.rs
```diff
@@ -0,0 +1,92 @@
...
+    let builder = http::response::Builder::new();
+    let http_resp = builder.status(web_sys_resp.status()).body(resp_body)?;
+    Ok(http_resp)
+}
+
+// FetchErrors are returned by both the rust and browser versions of fetch. Since
+// lower level errors in each case will be coming from two different places I implemented
+// FetchError as lossy of the error types underneath: it holds an error string.
+#[derive(Debug)]
+pub struct FetchError {
```
phritz, line 55: i know i can do this with `Enumtype(String)` without implementing Error and that elsewhere we are enumerating error cases as opposed to using strings to describe them. i did this first before i understood that. my sense is that if you'll want us to standardize around enumerated types rather than what we have here or FetchError(String).

src/embed/connection.rs
```diff
@@ -272,6 +274,18 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
     Ok(PutResponse {})
 }
 
+// This fn just forwards to sync::begin_sync at this point. execute() requires its
+// do_ers to take txns even if they don't need it so this fn also helps begin_sync
+// have a simpler signature.
+async fn do_begin_sync<'a, 'b>(
+    _store: &'a dag::Store,
```
aboodman, line 281: [style] Just name these params `_`. I believe that is idiomatic in Rust to say that something is not used, and in that case don't even need the comment.

src/embed/http.rs
```diff
@@ -0,0 +1,91 @@
...
+use std::error::Error;
+use std::fmt;
+use std::str;
+use wasm_bindgen::JsCast;
+use wasm_bindgen::JsValue;
+use wasm_bindgen_futures::JsFuture;
+use web_sys::{RequestInit, RequestMode};
+
+// browser_fetch makes and HTTP request over the network via the browser's Fetch API.
+// It is intended to be used in wasm; it won't work in regular rust. The response contains
```
aboodman, line 11: [idle speculation - no task for now] It probably makes more sense to turn `src/wasm.rs` into `src/wasm` and put this there along with some abstraction from the "core" to the correct version.

src/embed/http.rs
```diff
@@ -0,0 +1,91 @@
...
+) -> Result<http::Response<String>, FetchError> {
+    let mut opts = RequestInit::new();
+    opts.method(http_req.method().as_str());
+    opts.mode(RequestMode::Cors);
+    let js_body = JsValue::from_str(http_req.body());
+    opts.body(Some(&js_body));
+    let web_sys_req = web_sys::Request::new_with_str_and_init(&http_req.uri().to_string(), &opts)?;
+    let h = web_sys_req.headers();
+    for (k, v) in http_req.headers().iter() {
+        h.set(k.as_ref(), v.to_str().unwrap())?;
```
aboodman, line 29:
```bash
$ su phritz
$ echo "BUT THE PAAAANICS"
```

aboodman: (why not just make this one of the `FetchError`s?)

src/embed/http.rs
aboodman: I really like the enumerated types and think we could create machinery to make them easier to use. What do you think?

src/embed/sync.rs
```diff
@@ -0,0 +1,128 @@
+use super::http as embed_http;
```
aboodman, line 1: I thought you wanted to do `sync` as a peer to `db`, e.g., `src/sync/pull.rs`. I liked that plan.

src/embed/sync.rs
```diff
@@ -0,0 +1,128 @@
...
+    // TODO do something with the response
+    Ok(BeginSyncResponse {})
+}
+
+#[derive(Debug)]
+pub enum BeginSyncError {
+    PullFailed(PullError),
+}
+
+impl From<PullError> for BeginSyncError {
```
aboodman, line 25: I think it is better overall to not use these From impls (despite the recommendations online) because they are one-to-one with source error. Whereas in real code you'll frequently have multiple places in the same function that cause the same error type, and will want to give them different enumerated tags. So I think we should just do `.map_err` at the call site. It also saves all this impl goop.

src/embed/sync.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[allow(dead_code)]
+    last_mutation_id: String,
+    // 	TODO Patch          []kv.Operation `json:"patch"`
+    #[nserde(rename = "checksum")]
+    #[allow(dead_code)]
+    checksum: String,
+    // TODO ClientViewInfo ClientViewInfo `json:"clientViewInfo"`
+}
+
+pub async fn pull(begin_sync_req: &BeginSyncRequest) -> Result<PullResponse, PullError> {
```
aboodman, line 58: I'm super here for these bare functions.

src/embed/sync.rs
phritz: i perceived the embed::sync module as that kind of peer but happy to move it. should i think of embed as the embedded api, where the api implementations generally live elsewhere?

aboodman:
I guess there's really no difference it's just directory structure. But so
far we only have a single-level flat directory structure. Conceptually I
was thinking of everything inside of 'embed' having to do with embedding,
and since sync stands alone, I expected to see it outside of embed as its
own top level dir under src.

src/embed/http.rs
phritz: 1. can't decide whether i'm annoyed at having all the fidelity i don't care about versus reassured that each case is doing the right thing. i liked the go style of only formalizing that which the caller would wants to treat like a first class thing. but enumerating all cases maybe does seem more rusty. 
2. i like human-readable error strings that carry the bits of information i'm interested, when such bits exist. ParseError is a lot less useful than ParseError("unexpected token 'noll' attempting to parse as json '{"foo": noll}'"). so as long as we can have that sounds good.

i dont have a great deal of experience and the enumerated types seem pretty lightweight so i'm game for giving it a go.

src/embed/http.rs
phritz: panics indeed, nice catch. either i missed this on my sweep of cleaups or i had a type assertion thingy above, dunno.

aboodman:
On Tue, Aug 11, 2020 at 6:21 PM Phritz <notifications@github.com> wrote:

> *@phritz* commented on this pull request.
>
> In src/embed/http.rs
> <https://github.com/rocicorp/repc/pull/79#discussion_r468996277>:
>
> > +    let web_sys_resp: web_sys::Response = js_web_sys_resp.dyn_into().unwrap();
> +    let body_js_value = JsFuture::from(web_sys_resp.text()?).await?;
> +    let resp_body = body_js_value.as_string().ok_or(FetchError::new(
> +        "could not get http response body as string",
> +    ))?;
> +
> +    let builder = http::response::Builder::new();
> +    let http_resp = builder.status(web_sys_resp.status()).body(resp_body)?;
> +    Ok(http_resp)
> +}
> +
> +// FetchErrors are returned by both the rust and browser versions of fetch. Since
> +// lower level errors in each case will be coming from two different places I implemented
> +// FetchError as lossy of the error types underneath: it holds an error string.
> +#[derive(Debug)]
> +pub struct FetchError {
>
>
>    1. can't decide whether i'm annoyed at having all the fidelity i don't
>    care about versus reassured that each case is doing the right thing. i
>    liked the go style of only formalizing that which the caller would wants to
>    treat like a first class thing. but enumerating all cases maybe does seem
>    more rusty.
>
> My pitch for this bit has two parts:

a. We can probably <furious handwaving> automate this so that the fidelity
is generated -- you don't have to do anything to create it. If it happened
automatically would it still bother you? I kind of envision it being
amazing to hover an error type and see a tree of all the different error
cases.

b. I find it kind of stressful trying to figure out which error cases
caller should care about. Maybe that's just API design though.

>
>    1. i like human-readable error strings that carry the bits of
>    information i'm interested, when such bits exist. ParseError is a lot less
>    useful than ParseError("unexpected token 'noll' attempting to parse as json
>    '{"foo": noll}'"). so as long as we can have that sounds good.
>
> To be clear the values of the enum can be any Rust type. So you'd really
have something like:

FetchError(ResponseParseError(DeJsonError("unexpected token 'noll'
attempting to parse as json '{"foo": noll}'"))).

No information is lost. It's basically the same as Go's wrap idiom, except
that it's strongly typed.

src/embed/http.rs
aboodman: PS - I looked into this and I think the problem is somewhat lessened in Rust/web. Panics take down the wasm vm, but not the web page. So the webpage will still "work" modulo replicache (so probably not -- heh). You could also theoretically reload the wasm in that case.

OTOH, given the enumerated errors, I really feel a lot less sympathy for having any panics in the codebase at all, as I feel it is easy enough to handle them correctly.

src/embed/sync.rs
phritz: hmmm. in this case i would find the proliferation of error types and map_err everywhere to be goop that gets in the way of understanding what is happening. i like that the errors here are 1-1 -- most of the source errors have the same type so with one from i have handled almost all of them. and i can still tell as a developer what went wrong because i'm capturing their debug output, but i'm not forced to enumerate all the separate ways this thing can fail. 

i have a hard time reconciling the feeling that the enumerated types are good at the sync level but not good here. i realize it might be a contradiction but it still feels different. maybe because in some sense this is the lowest (rust) level? like this is the bottommost place that a coherent thing is happening (an http fetch); at any lower level the things that are happening re in service of this level, so it feels natural to bottom out here for some (admittedly maybe contradictory) reason.

hmmmm.

src/embed/sync.rs
phritz: > I guess there's really no difference it's just directory structure. But so
far we only have a single-level flat directory structure. Conceptually I
was thinking of everything inside of 'embed' having to do with embedding,
and since sync stands alone, I expected to see it outside of embed as its
own top level dir under src.

I'll move it in a future PR.

src/embed/sync.rs
aboodman: It seems pretty likely that somewhere in this sync code you'll have a function where http::Error happens twice in the same function no? It was exactly that kind of issue that led me away from this pattern.

aboodman: > not forced to enumerate all the separate ways this thing can fail.

This part is annoying with my proposal, I agree. I was hoping I could put something together like:

```
fetch(pull_url).map_err(FetchPullError)?
```

... and the FetchPullError would automatically go into an enum for the function. Then you wouldn't have to enumerate them, it would happen automatically.

phritz:
merging but let's keep the convo up there are still unresolved questions.

re automatically adding enumerated errors i'm not sure how to do that.

re having all errors enumerated like you say it obscures which ones you should really worry about, not a huge fan of that.

>  I kind of envision it being amazing to hover an error type and see a tree of all the different error cases.

i think it would be amazingly noisy, crowding out the few really important things with a wall of edge cases that can never happen in reality or that are not important. just imagine the exponential explosion of ways high level things can fail where 99% of them do not matter to you as a programmer but literally 1 out of those 100 is the thing that you are interested in and now you have to figure out what it is and find it.

aboodman:
On Tue, Aug 11, 2020 at 7:28 PM Phritz <notifications@github.com> wrote:

> merging but let's keep the convo up there are still unresolved questions.
>
👍

> re automatically adding enumerated errors i'm not sure how to do that.
>
Me either. I'm thinking of this as some kind of future enhancement. Macros
probably!

> re having all errors enumerated like you say it obscures which ones you
> should really worry about, not a huge fan of that.
>
> I kind of envision it being amazing to hover an error type and see a tree
> of all the different error cases.
>
> i think it would be amazingly noisy, crowding out the few really important
> things with a wall of edge cases that can never happen in reality or that
> are not important. just imagine the exponential explosion of ways high
> level things can fail where 99% of them do not matter to you as a
> programmer but literally 1 out of those 100 is the thing that you are
> interested in and now you have to figure out what it is and find it.
>
It won't be exponential. The errors a particular function returns is still
some small number - just the call sites inside that function that error.
Each of those errors will themselves perhaps be composed of other cases,
but you would only discover that as necessary.

It's really only a little different than what you are already doing. Take a
look at:

https://github.com/rocicorp/repc/pull/79/files#diff-ce832c9f1c0fa06a2932518e61d21504R104

You already have a union tagging each error that can happen from pull. That
would be exactly the same under my proposal. Same number of enum tags, in
fact same exact enum. The only differences would be:

1/ get rid of 15 lines of From impl boilerplate
2/ Add .map_err() suffix to each of the call sites that generate errors
3/ Now if a function has two callsites that return the same error type, no
problem

That's it.

I had the same worry as you at first that this would lead to a lot of noise
but when I tried it I actually ended up reassured that there actually
*aren't* that many unique error cases. Each function can only return a
handful, practically speaking.

phritz:
ok. fyi though fetcherrors are going to be a little different because there
are two separate sources for them: the rust http client and the browser
fetch api. but i will model them this way, just i guess with more string
inner types to hide the different sources.

tests/wasm.rs
```diff
@@ -255,3 +259,29 @@ async fn get_put() {
...
+// 2. wasm-pack test --chrome -- --test wasm
+// 3. open developer tools in a browser window
+// 4. navigate to 127.0.0.1:8000
+// 5. verify the request and response by inspection:
+//     - method
+//     - http headers
+//     - outgoing and incoming body
+//
+// #[wasm_bindgen_test]
+// async fn test_browser_fetch() {
```
arv, line 275: Can't you use a custom cfg or feature so we at least get compile errors?

phritz: Yes, that'd be much better. In fact all we have to do is comment out the wasm_bindgen_test attribute above instead of the whole function and add an allow dead code and we'll get compile errors. Will do that in my next follow up.

PR #80, opened 2020-08-12
add rust fetch and test

phritz:
- i used hyper for pull's http request. requirements in selecting a client were must be async, must be widely used, must support cancel/timeout, and must not pull in enormous deps like libcurl. hyper is relatively low level, but it still pulls in the tokio runtime, which is a bummer. since calls to hyper/tokio are not called when we compile to was (invocation is guarded by a const bool) they should be dead-code-eliminated, so this overhead should not be felt for web (need to verify this once we can get wasm to compile; see below). anyway, i'd love to debate hyper vs something else when it matters, which i contend is not now.

- hyper/tokio are networking code which does not compile for wasm (wasm-pack build fails). i do not think this is specific to hyper/tokio, i think *any* networking code that we attempted to include when building for wasm would cause it to fail. i need to find a way to exclude these dependencies when building for wasm, as they are not needed in that case.

progress towards #43


aboodman:
[workflow comment] in cases such as these where the commits in a PR are not intended to be logically distinct parts that can/should be reviewed separately, you can squash locally and `git push origin foo --force` to make that clear to reviewer.

src/embed/connection.rs
```diff
@@ -273,15 +273,19 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
...
-// do_ers to take txns even if they don't need it so this fn also helps begin_sync
-// have a simpler signature.
 async fn do_begin_sync<'a, 'b>(
     _: &'a dag::Store,
     _: &'b TxnMap<'a>,
     req: BeginSyncRequest,
 ) -> Result<BeginSyncResponse, sync::BeginSyncError> {
-    let begin_sync_response = sync::begin_sync(&req).await?;
+    // TODO move client up to process() or into a lazy static so we can re-use.
+    let hyper_client_owned: hyper::Client<hyper::client::HttpConnector>;
```
aboodman, line 282: [style] leave off `owned`, unless there is a borrowed variant somewhere in here.

src/embed/connection.rs
```diff
@@ -273,15 +273,19 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
...
+    let mut hyper_client_option = None;
```
aboodman, line 283: It seems like you can use match to condense this somewhat and avoid the mut option:

```
let hyper_client = match sync::USE_BROWSER_FETCH {
    true => Some(hyper::Client::New()),
    false => None,
}
```

If you then pass `&hyper_client` into `begin_sync` I believe that all the methods that get the thingy out of the option will give it to you by ref too (otherwise there are helpers for that).

src/embed/connection.rs
```diff
@@ -273,15 +273,19 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
...
+    if !sync::USE_BROWSER_FETCH {
+        hyper_client_owned = hyper::Client::new();
+        hyper_client_option = Some(&hyper_client_owned);
+    }
+    let begin_sync_response = sync::begin_sync(hyper_client_option, &req).await?;
```
aboodman, line 288: This is kind of cool -- I presume that http client is going to get passed all over the place to async functionality and it's totally fine because it's immutable.

src/embed/http.rs
```diff
@@ -1,96 +1,234 @@
...
+) -> Result<http::Response<String>, FetchError> {
+    let (parts, req_body) = http_req.into_parts();
+    let mut builder = hyper::Request::builder()
+        .method(parts.method.as_str())
+        .uri(&parts.uri.to_string());
+    for (k, v) in parts.headers.iter() {
+        builder = builder.header(
+            k,
+            v.to_str()
+                .map_err(|e| FetchError::InvalidRequestHeader(s(e)))?,
```
aboodman, line 33: [help] I don't understand why `s()` is needed. When I implemented this pattern myself, I found that a nice string was automatically generated, recursively, when I tried to print the top level error. So the functionality in `s` was not necessary.

Doing it this way you are discards other information in `e` -- its type, which is a big part of why I liked this pattern. I'm not in favor of it if we have to do this.

Perhaps this is where the `std::err::Error` trait comes in?

src/embed/http.rs
```diff
@@ -1,96 +1,234 @@
...
+
+    let mut hyper_resp = hyper_client
+        .request(hyper_req)
+        .await
+        .map_err(|e| FetchError::RequestFailed(s(e)))?;
+    let http_resp_builder = http::response::Builder::new();
+    let http_resp_bytes = hyper::body::to_bytes(hyper_resp.body_mut())
+        .await
+        .map_err(|e| FetchError::ErrorReadingResponseBody(s(e)))?;
+    let http_resp_string =
```
aboodman, line 48: [nit] I guess if we're always decoding JSON we do want this as a string, but I'd probably still type the `Response::body` as `Vec<u8>` and just let callers who want it as a string do that.

phritz:
> [workflow comment] in cases such as these where the commits in a PR are not intended to be logically distinct parts that can/should be reviewed separately, you can squash locally and git push origin foo --force to make that clear to reviewer.

We have had different defaults. As you point out, I usually squash on merge and don't have separate logical commits. I call it out in the PR when the commits *are* separately logical eg https://github.com/rocicorp/replicache-client/pull/93#issue-418786004. I think based on your comments I should instead always go with separate logical commits.

src/embed/http.rs
```diff
@@ -1,96 +1,234 @@
 use std::convert::TryFrom;
```
aboodman, line 1: It's great that in Replicache http is required only at this extremely high level. In Noms it is required very low down in the stack.

aboodman:
I guess actually now that I think of it the way you do it makes more
sense since reviewers can see the edit history in relation to their
comments. I will start doing it that way too.

Cargo.toml
```diff
@@ -12,24 +12,28 @@ benchmark = []
...
 async-trait = "0.1.36"
+bytes = "0.5.6"
 console_log = { version = "0.2" }
 console_error_panic_hook = { version = "0.1.1", optional = true }
 data-encoding = "1.1.1"
 flatbuffers = "0.6.1"
 futures = "0.3.5"
 http = "0.2.1"
+# hyper implies tokio below. Should be dead-code-removed in wasm.
+hyper = "0.13"
```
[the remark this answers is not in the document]
phritz: > It's weird that it doesn't pull it in itself? Is that because the runtime it uses is meant to be pluggable?

We depend directly on it here https://github.com/rocicorp/repc/pull/80/files#diff-0de8ae7a07d52b7410d27f30e00a32dbR75

> I'm a bit concerned about pulling in tokio ... because it seems like it overlaps with functionality provided by async-std, which we've already chosen. Will that cause problems?

I don't know.

I didn't think we were even using the async_std runtime but it looks like I was mistaken: https://github.com/rocicorp/repc/blob/d4f67ab80ddb33b7556055c14ec7be4e56d1e2fc/src/kv/idbstore.rs#L3. I'm a little unsure about that because I'm not sure it's possible to prevent it from spawning threads without explicit configuration, and it has some magic to "detect blocking" (look for that phrase in https://async.rs/blog/stop-worrying-about-blocking-the-new-async-std-runtime/). I don't see us setting ASYNC_STD_THREAD_COUNT anywhere. I think @whiten said in slack he is looking at this as a potential cause of a problem, possibly related to the next thing I think.

We are primarily using the wasm_bindgen runtime via https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen_futures/fn.spawn_local.html. This should work both in wasm and in regular rust I think. 

Regarding tokio's runtime, I *thought* that we we weren't using it because we are not using the attribute "#[tokio::main]" or anything else that obviously configures or uses the runtime like tokio's task::spawn. I assumed that we were just creating the http request future and then when we await it it's all running on our main thread. However I looked under the hood into the http client and indeed it is using a tokio executor and tokio tasks. I have no idea what kinds of problems to expect if we are both spawning tasks with wasm_bindgen and with tokio. I know that in principle you can plug other runtimes into tokio and have it work, but no idea how hard that would be in this case. However the good news is that this potential problem is not present in wasm, just in not-wasm and and thus potentially in integration tests.

I think we should spin the runtime question out into an issue.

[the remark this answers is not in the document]
phritz: https://github.com/rocicorp/repc/issues/82

src/embed/connection.rs
phritz: yes it is seeming like inner mutability is an important thing for shared state

src/embed/http.rs
phritz: this was an explicit choice i call out in the comment bc there were multiple lower level types. i agree that using std::err:error is better, not sure why i didn't do that. anyway created a follow up: https://github.com/rocicorp/repc/issues/83

src/embed/http.rs
phritz: 100% of our callers presently want a String so going to leave it as is

phritz:
I believe I have addressed most of the comments here. I spun off the following so I can make progress:
https://github.com/rocicorp/repc/issues/82
https://github.com/rocicorp/repc/issues/83

If you are curious how the conditional compilation works search for "cfg" in this commit: https://github.com/rocicorp/repc/pull/80/commits/2388f90dec7fb9c1abecb399c898bb1f68bfd7ac

I'm going to ignore the question of whether we should use these three lines or those four lines to express this thing that everyone understands both ways as something we have already spent too much time on, too much being more than zero.

PR #84, opened 2020-08-14
uncomment but dont run wasm fetch test

phritz:


phritz:
@arv ask and ye shall receive

PR #92, opened 2020-08-14
Plumb transaction name/args through to stored commit.

aboodman:
Provided as two separate commits for your reviewing pleasure.


aboodman:
No the args are currently passed through embedded in the outer payload. We could change the protocol, but I was trying to stay as close to it as possible.

Issue #94, opened 2020-08-14
Plumb the rest of the commit metadata for local through from dispatch


aboodman:
After #92 we will still have the following bits we need to capture and store:

- mutation_id
- original_hash

PR #96, opened 2020-08-15
Implement mutationID generation

aboodman:
Toward #94

### DIFF
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -105,6 +105,18 @@ impl Commit {
         self.commit().value_hash().unwrap()
     }
 
+    pub fn next_mutation_id(&self) -> u64 {
+        let meta = self.commit().meta().unwrap();
+        let mid = match meta.typed_type() {
+            commit::MetaTyped::LocalMeta => meta.typed_as_local_meta().unwrap().mutation_id(),
+            commit::MetaTyped::SnapshotMeta => {
+                meta.typed_as_snapshot_meta().unwrap().last_mutation_id()
+            }
+            commit::MetaTyped::NONE => unreachable!(),
+        };
+        mid + 1
+    }
+
     fn validate(buffer: &[u8]) -> Result<(), LoadError> {
         use LoadError::*;
         let root = commit::get_root_as_commit(buffer);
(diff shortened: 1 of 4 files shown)


src/db/commit.rs
```diff
@@ -448,29 +460,31 @@ mod tests {
...
-                assert_eq!(sm.server_state_id(), "server_state_id");
+                assert_eq!(sm.last_mutation_id(), 2);
+                assert_eq!(sm.server_state_id(), "server_state_id 2");
             }
         }
-        assert_eq!(local.meta().local_create_date(), "local_create_date");
-        assert_eq!(local.meta().basis_hash(), Some("basis_hash"));
-        assert_eq!(local.meta().checksum(), "checksum");
-        assert_eq!(local.value_hash(), "value_hash");
+        assert_eq!(snapshot.meta().local_create_date(), "local_create_date 2");
```
aboodman, line 483: whoops

src/db/commit.rs
```diff
@@ -105,6 +105,18 @@ impl Commit {
         self.commit().value_hash().unwrap()
     }
 
+    pub fn next_mutation_id(&self) -> u64 {
+        let meta = self.commit().meta().unwrap();
```
phritz, line 109: Not something that came in with this PR but: the highly overloaded nature of "commit" in this code I found super confusing. commit()'s signature is basically the following, which is hard to make heads or tails of:
```
struct Foo {}

impl Foo {
  fn foo() -> foo::Foo
}
```
Can we bring `super::commit_generated::commit` in as `commit_fb` or similar and call `commit()` something like `fb()` so the reader can see what is what?

src/db/commit.rs
```diff
@@ -105,6 +105,18 @@ impl Commit {
...
+        let mid = match meta.typed_type() {
+            commit::MetaTyped::LocalMeta => meta.typed_as_local_meta().unwrap().mutation_id(),
```
phritz, line 111: Do you have creating a genesis snapshot when we create the db on your list of todos?

src/db/commit.rs
aboodman: That's a good idea. https://github.com/rocicorp/repc/issues/100

src/db/commit.rs
aboodman: Nope! https://github.com/rocicorp/repc/issues/102

Issue #77, opened 2020-08-08
repc testing


phritz:
Working on sync brings to light some questions about how we are and should be testing repc.

**Assumptions/background:**
- http requests are an important part of the sync protocol
- as far as I can tell, by design we cannot easily run a web server when compiling to wasm in the browser (no sockets, eg). OK, sure, I can imagine some sort of http-over-websocket gymnastics, but that doesn't sound _easy_: something that works well off the shelf and doesn't make your head explode. This means that testing sync which requires http requests is necessarily going to be limited in the wasm-in-browser case because we can't run a web server. Mocking is of course an option.
- wasm-in-the-browser is not the only platform we're going to want to run on, for _example_ iOS and Android are enormous markets
- given the way we have architected repc, the main differences across platforms seem likely to be 1) how http requests are made and 2) the kv store implementation we use, and maybe 3) parameters eg for data structures. (Are there others I'm missing?).
- the diffserver is written in go, as is the [sample data layer](https://github.com/rocicorp/replicache-sample-todo/tree/master/serve).
- we derived a lot of benefit from having [integration tests](https://github.com/rocicorp/replicache-client/blob/master/repm/sync_integration_test.go#L332) in replicache-client that exercise the whole stack under the bindings. The integration tests instantiate a real client, a real diffserver, and a fake data layer all in process and we test the protocol using the client's dispatch api. I think the only differences between the integration tests and real life are A) client is driven by the tests, not the bindings and B) the data layer is fake and C) the components are all running in the same process. Otherwise it's like the real thing (real http requests to real http servers, etc). 
- the [account service](https://github.com/rocicorp/account-service/) is written in go and is composed of a CLI that talks rpc to lambdas backed by an amazon RDS db. The account service is not yet complete and is not integrated with the client (eg to use account-service-generated account ids) or diffserver (eg to use account-service-managed data layer urls or to keep track of diffserver fetches by account if we do that). The account service has [integration tests](https://github.com/rocicorp/account-service/blob/6e94bfb58676f50d669313c132498376a4064354/integration_test/cli_test.go#L414) that drive the CLI talking to a real in-process account service server and a fake in process github OAUTH server. _If we decided it was worth the cost_, we could pretty easily include the account service together with the diffserver/fake data layer/replicache-client in integration tests. 

**Discussion: How to test sync**

Given that we can't instantiate a web server in wasm we have to choose how to test sync. We could 
[...]

PR #104, opened 2020-08-17
Add FFI interface to allow dispatch() from Go.

whiten:
For sync integration testing, #77.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ wasm-bindgen-test = "0.3.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 bytes = "0.5.6"
+env_logger = "0.7.1"
 hyper = "0.13" # Implies tokio.
 tokio = { version = "0.2", features = ["full"] } # For hyper.
 
@@ -65,7 +66,7 @@ features = [
 ]
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["staticlib", "cdylib", "rlib"]
 
 [profile.release]
 codegen-units = 1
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -65,7 +66,7 @@ features = [
 ]
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["staticlib", "cdylib", "rlib"]
```
aboodman, line 69: lol. Maybe worth documenting what each is needed for.

src/ffi.rs
```diff
@@ -0,0 +1,159 @@
...
+
+pub fn init_log() {
+    INIT.call_once(env_logger::init);
+}
+
+#[no_mangle]
+pub extern "C" fn dispatch(
+    db_name: *const c_char,
+    rpc: *const c_char,
+    args: *const c_char,
```
aboodman, line 17: Do we need to worry about embedded nulls in the args? I feel like we do.

whiten: By worry, do you mean "don't crash if present", or "retain through all transformations"? I think we expect the first two arguments to be regular strings and the third to be utf-8 encoded JSON, so I presume it's just the former, right?

I think that's covered through the bad encoding tests like https://github.com/rocicorp/repc/pull/104/files#diff-49a4b4d69f811d5619c481802227b86eR127.

Cargo.toml
whiten: Yeah, fair. Let me look up some good descriptions. I'm also curious if we can compile only some of these for various modes, since we'll never need a staticlib for wasm32, for instance.

src/ffi.rs
aboodman: I mean retain so that we can properly support client expectations.

I forgot that utf8 doesn't have nulls in it. I think all three should be defined to be utf8, so hooray.

PR #108, opened 2020-08-19
Various changes to db::Commit API

### LINKED ISSUE #107: additional small changes/qs re sync
I looked over what we need to implement sync beyond what is represented in issues already and came up with the small grab bag of issues/questions below, might as well discuss up front.

1. We need to do a lot of traversing chains of commits for sync. This requires reading commits that are not at the head of their chain. I plan to make some small tweaks to the commit and tx interfaces to make this easier. Right now functions like `Commit::from_head` bundle two things together, getting the hash the named head is set to and then reading it into a Commit. I'm going to make at least the reading into a commit into a first class citizen. Something like the following, presumably this is uncontroversial:
   - let's have "load" mean "read from the store" so rename `Commit::load()` to `Commit::from_chunk()`.
   - add `Commit::load(hash, dag_read) -> Commit` that calls `dag_read.get_chunk(hash)` and then `Commit::from_chunk` on the result
   - have `from_head` call `dag_read.get_head` as it does today then `Commit::load`
1. Do we use multiple sync branches, allowing multiple syncs to run in parallel? IF enabling parallel syncs is important then i think we should make a small change to the protocol (see below) even though we can technically do it without the change. IF enabling parallel syncs is not important then it's even easier (see also below).
   - with no changes to the protocol the most straightforward thing is to not allows syncs in parallel. If we use a single named head to track the sync chain then if a new sync starts while one is already running it effectively cancels the old sync. (i can go into details if you want, not sure they are important.)
   - with no changes to the protocol we can support multiple parallel syncs with a creative use of named heads (a sync chain has a named head with name equal to the hash of its head commit; to extend a sync chain with a commit delete the old named head and crate a new named head with name hash of the new commit. sync chains wont be gc'd and we always have synchead/basis passed in in open/committransaction and maybeendsync.)
   - if we added syncid to the rebaseopts then we can enable multiple parallel syncs using a named head equal to the syncid. 
1. We need to add a way to remove a named head to clean up after sync. This seems straightforward.
1. We need to fix this bug https://github.com/rocicorp/repc/blob/e803f9a3b38da5bcce0112b3436d746ded9200b4/src/embed/connection.rs#L182 (basis is a hash, not a named head).

aboodman:
A series of independent commits that makes the following changes:

- Replace Commit::from_head with from_hash. There won't be a from_head convenience at this layer anymore.
- Remove db::Write::new_from_head
- Add db::Write::new_local and new_snapshot both of which can be constructed from either head name or hash
- Insert genesis commit at startup

Fixes #107, #102 , #97

### DIFF
--- a/src/db/read.rs
+++ b/src/db/read.rs
@@ -2,13 +2,19 @@ use super::commit::{Commit, FromHashError};
 use crate::dag;
 use crate::prolly;
 
+pub enum Whence {
+    Head(String),
+    #[allow(dead_code)]
+    Hash(String),
+}
+
 pub struct OwnedRead<'a> {
     dag_read: dag::OwnedRead<'a>,
     map: prolly::Map,
 }
 
 #[derive(Debug)]
-pub enum NewReadFromHeadError {
+pub enum ReadCommitError {
     UnknownHead(String),
     GetHeadError(dag::Error),
     CommitFromHeadError(FromHashError),
@@ -16,29 +22,39 @@ pub enum NewReadFromHeadError {
 }
 
 impl<'a> OwnedRead<'a> {
-    pub async fn new_from_head(
-        head_name: &str,
+    pub async fn from_whence(
+        whence: Whence,
         dag_read: dag::OwnedRead<'a>,
-    ) -> Result<OwnedRead<'a>, NewReadFromHeadError> {
-        use NewReadFromHeadError::*;
-        let read = dag_read.read();
-        let hash = read
-            .get_head(head_name)
-            .await
-            .map_err(GetHeadError)?
-            .ok_or(UnknownHead(head_name.to_string()))?;
-        let commit = Commit::from_hash(&hash, read)
-            .await
-            .map_err(CommitFromHeadError)?;
-        let map = prolly::Map::load(commit.value_hash(), dag_read.read())
-            .await
-            .map_err(MapLoadError)?;
+    ) -> Result<OwnedRead<'a>, ReadCommitError> {
+        let (_, _, map) = read_commit(whence, &dag_read.read()).await?;
         Ok(OwnedRead { dag_read, map })
     }
 
     pub fn as_read(&'a self) -> Read<'a> {
         Read::new(self.dag_read.read(), &self.map)
     }
+}
+
+pub async fn read_commit(
+    whence: Whence,
+    read: &dag::Read<'_>,
+) -> Result<(String, Commit, prolly::Map), ReadCommitError> {
+    use ReadCommitError::*;
+    let hash = match whence {
+        Whence::Hash(s) => s,
+        Whence::Head(s) => read
+            .get_head(&s)
+            .await
+            .map_err(GetHeadError)?
+            .ok_or_else(|| UnknownHead(s.to_string()))?,
+    };
+    let commit = Commit::from_hash(&hash, read)
+        .await
+        .map_err(CommitFromHeadError)?;
+    let map = prolly::Map::load(commit.value_hash(), read)
+        .await
+        .map_err(MapLoadError)?;
+    Ok((hash, commit, map))
 }
 
 pub struct Read<'a> {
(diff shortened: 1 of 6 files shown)


src/db/read.rs
```diff
@@ -90,22 +90,21 @@ mod tests {
...
     use str_macro::str;
 
     #[async_std::test]
     async fn basics() {
-        let kv = MemStore::new();
-        init_db(&kv, "main", "local_create_date").await.unwrap();
-        let kvw = kv.write().await.unwrap();
-        let dw = dag::Write::new(kvw);
+        let ds = dag::Store::new(Box::new(MemStore::new()));
+        init_db(ds.write().await.unwrap(), "main", "local_create_date")
```
aboodman, line 99: I had to change the signature of `init_db` to be compatible with objects `connection.rs` has available (pass in a dag::Write instead of kv::Store).

src/db/write.rs
```diff
@@ -31,14 +30,18 @@ pub struct Write<'a> {
     meta: Meta,
 }
 
+#[derive(Debug)]
```
aboodman, line 33: This wasn't completely necessary but nice for consistency, and if other errors come up during further dev of InitDB.

src/db/read.rs
```diff
@@ -36,6 +35,28 @@ impl<'a> OwnedRead<'a> {
     }
 }
```
phritz, line 37: can you comment the return types? (also: we should consider having `type Hash = String` and `type HeadName = String` for clarity).
```
// read_commit reads a commit specified by whence from the dag and returns its hash, Commit, and Map.
```

src/db/read.rs
```diff
@@ -36,6 +35,28 @@ impl<'a> OwnedRead<'a> {
...
 
+pub async fn read_commit(
```
phritz, line 38: Feedback here is similar to the feedback about `from_head()`: I wish this function did less, or that there was another function that just returned the hash and commit. Returning the hash makes sense bc the commit might be from a head. But what if i don't want the map because i'm just hunting for a commit with a particular last mutation id?

I think we should prefer small, orthogonal api calls that each do one thing and then let the caller compose. (Or if it is a common case eg here to also want the map with the commit then we can provide a free fn like this.)

src/db/read.rs
```diff
@@ -36,6 +35,28 @@ impl<'a> OwnedRead<'a> {
...
+    read: &dag::Read<'_>,
+) -> Result<(String, Commit, prolly::Map), ReadCommitError> {
+    use ReadCommitError::*;
+    let hash = match whence {
+        Whence::Hash(s) => s,
+        Whence::Head(s) => read
+            .get_head(&s)
+            .await
+            .map_err(GetHeadError)?
+            .ok_or_else(|| UnknownHead(s.to_string()))?,
```
phritz, line 49: i have to admit the error mapping and ? do make the code a lot more readable than go

src/db/read.rs
```diff
@@ -91,9 +113,10 @@ mod tests {
             .await
             .unwrap();
 
-        let kvr = kv.read().await.unwrap();
-        let dr = dag::OwnedRead::new(kvr);
-        let r = OwnedRead::new_from_head("main", dr).await.unwrap();
+        let dr = ds.read().await.unwrap();
+        let r = OwnedRead::from_whence(Whence::Head(str!("main")), dr)
```
phritz, line 117: maybe we should be using the constant i saw somewhere instead of "main"

src/db/read.rs
aboodman: This function is internal to the module, it's an implementation detail of read.rs and write.rs.

phritz: ah, i expected a way to read from head or hash on Commit and saw it was pub and assumed this was the thing i should use to traverse the chain. instead i think I should use Read::get_head to get the head hash then Commit::from_hash to read the commit?

PR #111, opened 2020-08-20
add tests for pull

phritz:
two logical commits

### DIFF
--- a/src/embed/sync.rs
+++ b/src/embed/sync.rs
@@ -16,8 +16,9 @@ pub async fn begin_sync(
         base_state_id: "TODO".to_string(),
         checksum: "TODO".to_string(),
     };
-    let sync_id = "TODO";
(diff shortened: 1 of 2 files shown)


src/embed/sync.rs
```diff
@@ -10,7 +10,22 @@ pub async fn begin_sync(
     fetch_client: &fetch::client::Client,
     begin_sync_req: &BeginSyncRequest,
```
aboodman, line 11: Is there a reason to pass this by reference? It seems like you can avoid cloning data_layer_auth below if you take ownership of this param.

aboodman:
LGTM

src/embed/sync.rs
phritz: done in https://github.com/rocicorp/repc/pull/193

PR #112, opened 2020-08-21
add base_snapshot helper

phritz:
No test for begin_sync yet, that comes next.

### DIFF
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -180,6 +180,22 @@ impl Commit {
         let chunk = dag::Chunk::new(builder.collapse(), &[value_hash]);
         Commit { chunk }
     }
+
+    pub async fn base_snapshot(
+        hash: &str,
+        dag_read: &dag::Read<'_>,
+    ) -> Result<Commit, FromHashError> {
+        use FromHashError::*;
+        let mut commit = Commit::from_hash(hash, dag_read).await?;
+        while !commit.meta().is_snapshot() {
+            let meta = commit.meta();
+            let basis_hash = meta
+                .basis_hash()
+                .ok_or_else(|| ChunkMissing("None".to_string()))?;
+            commit = Commit::from_hash(basis_hash, dag_read).await?;
+        }
+        Ok(commit)
+    }
 }
 
 pub struct Meta<'a> {
@@ -208,6 +224,13 @@ impl<'a> Meta<'a> {
                 fb: self.fb.typed_as_snapshot_meta().unwrap(),
             }),
             commit::MetaTyped::NONE => panic!("notreached"),
+        }
+    }
+
+    pub fn is_snapshot(&self) -> bool {
+        match self.typed() {
+            MetaTyped::Local(_) => false,
+            MetaTyped::Snapshot(_) => true,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


phritz:
@aboodman maybe you can tell me how i can delete the lifetime parameter on begin_sync? clippy told me i could and i was surprised, as seems to be necessary at the upper levels (eg, connection).

src/db/commit.rs
```diff
@@ -180,6 +180,22 @@ impl Commit {
         let chunk = dag::Chunk::new(builder.collapse(), &[value_hash]);
         Commit { chunk }
     }
+
+    pub async fn base_snapshot(
+        hash: &str,
+        dag_read: &dag::Read<'_>,
+    ) -> Result<Commit, FromHashError> {
```
aboodman, line 187: Define a separate enum containing errors for this function: BaseSnapshotError. Instead of ChunkMissing use something like NoBasisHash(String) where the String parameter is the hash of the chunk lacking a basis.

src/embed/sync.rs
```diff
@@ -1,22 +1,47 @@
...
     fetch_client: &fetch::client::Client,
     begin_sync_req: &BeginSyncRequest,
 ) -> Result<BeginSyncResponse, BeginSyncError> {
+    use BeginSyncError::*;
+
+    let read = store.read().await.map_err(DbError)?;
+    let base_snapshot = Commit::base_snapshot(
+        &read
+            .read()
+            .get_head("main")
```
aboodman, line 23: For `"main"` use the constant that is defined in embed.rs or move it elsewhere.

src/embed/sync.rs
```diff
@@ -1,22 +1,47 @@
...
 use std::fmt::Debug;
 
 pub async fn begin_sync(
+    store: &dag::Store,
     fetch_client: &fetch::client::Client,
     begin_sync_req: &BeginSyncRequest,
 ) -> Result<BeginSyncResponse, BeginSyncError> {
+    use BeginSyncError::*;
+
+    let read = store.read().await.map_err(DbError)?;
```
aboodman, line 19: Give this its own entry in the error enum - `ReadError(dag::Error)`

src/embed/sync.rs
```diff
@@ -1,22 +1,47 @@
...
+    let base_snapshot = Commit::base_snapshot(
+        &read
+            .read()
+            .get_head("main")
+            .await
+            .map_err(DbError)?
```
aboodman, line 25: `GetHeadError`

src/embed/sync.rs
```diff
@@ -1,22 +1,47 @@
...
+            .unwrap(),
+        &read.read(),
+    )
+    .await
+    .map_err(CommitLoadError)?;
```
aboodman, line 30: `BaseSnapshotError`

src/db/commit.rs
```diff
@@ -277,10 +300,153 @@ pub enum FromHashError {
     LoadCommitFailed(LoadError),
 }
 
+#[cfg(test)]
```
aboodman, line 303: Why not put all this stuff in the `test` mod defined below?

src/db/commit.rs
```diff
@@ -277,10 +300,153 @@ pub enum FromHashError {
...
+        add_snapshot(&mut chain, &store).await;
+        let base_hash = store
+            .read()
+            .await
+            .unwrap()
+            .read()
+            .get_head("main")
+            .await
+            .unwrap()
+            .unwrap();
```
aboodman, line 423: this deserves some kind of award.

phritz:
(have not forgotten about these review comments)

src/db/commit.rs
phritz: i broke it out into a separate test-only module instead

src/db/commit.rs
phritz: do you think we should adjust our fmt parameters to allow longer lines? you can barely follow the flow of the function when a single let statement is spread over ten vertical lines.

aboodman:
I would definitely be in favor of adjusting our format rules *somehow*. I don't think it's the line-length rule that is causing the behavior you see there, or at least not completely. I agree it is illegible.

aboodman:
I forgot you can adjust it.

PR #116, opened 2020-08-25
wire pull into the db and add tests

phritz:


src/dag/write.rs
```diff
@@ -28,11 +28,14 @@ impl<'a> Write<'_> {
...
     }
 
-    pub async fn set_head(&mut self, name: &str, hash: &str) -> Result<()> {
-        Ok(self
-            .kvw
-            .put(&Key::Head(name).to_string(), hash.as_bytes())
-            .await?)
+    pub async fn set_head(&mut self, name: &str, hash: Option<&str>) -> Result<()> {
+        match hash {
+            None => Ok(self.kvw.del(&Key::Head(name).to_string()).await?),
```
aboodman, line 33: Dammit, I write this same bug every time I work on a system like this.

aboodman: We could also do `del_head`? What's more natural for caller code?

phritz: when it comes to utilities i often prefer to have smaller caller-composable pieces

PR #117, opened 2020-08-25
address code review comments

phritz:
comments given on https://github.com/rocicorp/repc/pull/112

### DIFF
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -3,6 +3,8 @@ use crate::dag;
 use flatbuffers::FlatBufferBuilder;
 use str_macro::str;
 
+pub const DEFAULT_HEAD_NAME: &str = "main";
+
 // Commit is a thin wrapper around the Commit flatbuffer that makes it
 // easier to read and write them. Commit::load() does validation
 // so that users don't have to worry about missing fields.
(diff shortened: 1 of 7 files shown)


src/db/commit.rs
```diff
@@ -185,15 +187,19 @@ impl Commit {
...
+        let mut commit = Commit::from_hash(hash, dag_read)
+            .await
+            .map_err(NoSuchCommit)?;
         while !commit.meta().is_snapshot() {
             let meta = commit.meta();
             let basis_hash = meta
                 .basis_hash()
-                .ok_or_else(|| ChunkMissing("None".to_string()))?;
-            commit = Commit::from_hash(basis_hash, dag_read).await?;
+                .ok_or_else(|| NoBasis(format!("Commit {} has no basis", commit.chunk.hash())))?;
```
aboodman, line 199: Nit, but I do not think the formatted string is necessary here. You could just have the parameter be the hash of the chunk with no basis. The name of the type tells you what the problem is, and when the debug formatter gets its hand on it, it prints something like `NoBasis(<hash>)` by default. Not requesting a change here, just a stylistic suggestion for next time. All this error handling is evolving anyway.

src/db/commit.rs
```diff
@@ -3,6 +3,8 @@ use crate::dag;
 use flatbuffers::FlatBufferBuilder;
 use str_macro::str;
 
+pub const DEFAULT_HEAD_NAME: &str = "main";
```
aboodman, line 6: I wasn't clear on this in my last review comment - my mistake. I think it makes more sense to leave this at the db layer and reference it from embed. I don't think Commit should know what the "default head name" is - it's a feature that it is ignorant of it. This is a micronit and I wasn't clear in my last review, so I will fix it myself, someday :).

PR #118, opened 2020-08-26
add checksum

### LINKED ISSUE #89: implement checksum
background: https://github.com/rocicorp/replicache-client/issues/31

basically the diffserver has a map and the client has a map, and during sync the diffserver delivers a patch to the client to apply to their map. to ensure we detect it if the maps get out of sync we compute a checksum over the map. this is merely a function of keys and values and not of any impl details of how they are stored. this checksum is stored in the commit https://github.com/rocicorp/repc/blob/3678fa5b5fbcce0ed061955ebddafa790565017c/src/db/commit.fbs#L29 and comes down in the response https://github.com/rocicorp/diff-server/blob/37135a25431fcd768c82279c9af35e9637b0fac5/serve/types/types.go#L20 https://github.com/rocicorp/repc/blob/3678fa5b5fbcce0ed061955ebddafa790565017c/src/embed/sync.rs#L52. the computation is here: https://github.com/rocicorp/diff-server/blob/master/kv/checksum.go. 

punt on json canonicalization for now.

we need to 
- implement checksum in rust
- file an issue in repc about json canonicalization. the diffserver uses https://godoc.org/github.com/gibson042/canonicaljson-go. i *think* we need the same thing in repc for correctness but maybe we can get away without it. anyway, let's not worry about it rn, just file an issue when the impl is complete.

phritz:
progress towards #89

### DIFF
--- a/src/checksum/mod.rs
+++ b/src/checksum/mod.rs
@@ -1,0 +1,161 @@
+use crc::{crc32, Hasher32};
+use std::fmt;
+use std::num::ParseIntError;
+use std::str::FromStr;
+
+// Checksum is a fast, incrementally computable, non-cryptographic
+// checksum of the contents of a map of bytes to bytes. The checksum
+// is commutative (it is not sensitive to the order of adds/removes)
+// so two maps with the same contents will have the same checksum no
+// matter in what order the keys were added (or removed/replaced).
+#[derive(PartialEq)]
+pub struct Checksum {
+    v: u32,
+}
+
+impl Checksum {
+    pub fn new() -> Checksum {
+        Checksum { v: 0 }
+    }
+
+    // Note: do not call add() if key is already in the map. Instead, call
+    // replace().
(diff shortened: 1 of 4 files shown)


src/checksum/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl fmt::Display for Checksum {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:08x}", self.v)
+    }
+}
+
+impl fmt::Debug for Checksum {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:08x}", self.v)
```
aboodman, line 54: Nit: Delegate one to the other to self-document they are the same? I think you can say something like:

```
fmt::Display::fmt(self, f)
```

phritz: done in https://github.com/rocicorp/repc/pull/193

PR #122, opened 2020-08-28
keep track of checksum in db::Write

phritz:


src/checksum/mod.rs
```diff
@@ -8,7 +9,7 @@ use std::str::FromStr;
 // is commutative (it is not sensitive to the order of adds/removes)
 // so two maps with the same contents will have the same checksum no
 // matter in what order the keys were added (or removed/replaced).
-#[derive(PartialEq)]
+#[derive(Clone, Copy, PartialEq)]
```
aboodman, line 12: I do not think that you need Clone. Remove this and remove all the calls to clone() in this PR. The Copy trait should change those assignments into copies automatically.

src/db/commit.rs
```diff
@@ -121,6 +123,7 @@ impl Commit {
         meta.local_create_date().ok_or(MissingLocalCreateDate)?;
         // basis_hash is optional -- the first commit lacks a basis
         meta.checksum().ok_or(MissingChecksum)?;
+        Checksum::from_str(meta.checksum().unwrap()).map_err(|_| InvalidChecksum)?;
```
aboodman, line 126: Nit: eliminate the scary looking unwrap by storing the result of the previous statement and using that in place of `meta.checksum().unwrap()`.

src/db/write.rs
```diff
@@ -106,16 +116,28 @@ impl<'a> Write<'a> {
     }
 
     pub fn put(&mut self, key: Vec<u8>, val: Vec<u8>) {
+        match self.map.get(&key) {
```
aboodman, line 119: [minifix] Internally map knows if the put() replaced an item (because its backing store returns that information from put). Change map.put() to return that to avoid the read here.

src/db/write.rs
```diff
@@ -106,16 +116,28 @@ impl<'a> Write<'a> {
...
+            Some(old_val) => self.checksum.replace(&key, old_val, &val),
+        };
         self.map.put(key, val)
     }
 
+    pub fn del(&mut self, key: Vec<u8>) {
+        let old_val = self.map.get(&key);
+        match old_val {
+            None => {}
+            Some(old_val) => self.checksum.remove(&key, old_val),
```
aboodman, line 130: Same thing we can easily have remove return whether the delete found anything -- the backing store knows.

aboodman, line 130: [style]

```
if let Some(old_val) = self.map.get(&key) {
   ...
}
```

... avoid the awkward empty None branch.

aboodman:
https://github.com/rocicorp/repc/pull/122/files#diff-ce832c9f1c0fa06a2932518e61d21504R310 ?

phritz:
well now, fuck me

phritz:
> > > Funny thing: there aren't any calls to clone in this pr! Totally news to me but apparently Copy requires Clone: https://doc.rust-lang.org/std/marker/trait.Copy.html#whats-the-difference-between-copy-and-clone. In any case the compiler complained without it.
> > https://github.com/rocicorp/repc/pull/122/files#diff-ce832c9f1c0fa06a2932518e61d21504R310 ?
>  well now, fuck me

No, I was right the first time: there are no clones. That line is a string in a unit test. Checksums are Strings in rpcs.

PR #126, opened 2020-08-30
Add getRoot RPC

aboodman:


src/embed/connection.rs
```diff
@@ -60,6 +57,7 @@ async fn connection_future<'a, 'b>(
         Some(v) => v,
     };
     match req.rpc.as_str() {
+        "getRoot" => execute(do_get_root, store, txns, req).await,
```
phritz, line 60: [cbb] no need to go back and address but so many places i see lists growing that are not in alpha order. makes it easier to find the thing if they are in order.

tests/wasm.rs
```diff
@@ -265,6 +265,24 @@ async fn get_put() {
     assert_eq!(dispatch(db, "close", "").await.unwrap(), "");
 }
 
+#[wasm_bindgen_test]
+async fn get_root() {
+    let db = &random_db();
+    assert_eq!(dispatch(db, "open", "").await.unwrap(), "");
+    assert_eq!(
```
phritz, line 272: if it's empty do we get ""? wait -- is it even possible for it to be empty in practice? i guess not, nm.

phritz (commented):
LGTM

src/embed/connection.rs
aboodman: good point, will alpha next time i'm there.

tests/wasm.rs
aboodman: This test is basically testing the behavior if we call `getRoot` after `open` as callers are supposed to do. I guess it is a valid question and worth testing what happens if you call `getRoot` before open.

PR #128, opened 2020-08-30
Swith from std::Mutex to async_std::Mutex.

aboodman:
This was causing a panic when contending on the mutex on the same thread in the sample app (we call openTransaction twice in a row in the same message pump).


phritz (commented):
LGTM

PR #130, opened 2020-08-31
implement MaybeEndSync

phritz:


phritz:
@aboodman can you look at how I did Mutation.args? Will that work? If so why didn't I just do that for patch value instead of stringifying it? Sorry I lost the thread on that.

src/sync/mod.rs
```diff
@@ -139,6 +148,114 @@ pub enum BeginSyncError {
...
+) -> Result<MaybeEndSyncResponse, MaybeEndSyncError> {
+    use MaybeEndSyncError::*;
+
+    // Ensure sync head is what the caller thinks it is.
+    let mut dag_write = store.write().await.map_err(WriteError)?;
+    let dag_read = dag_write.read();
+    let sync_head_hash = dag_read
+        .get_head(SYNC_HEAD_NAME)
+        .await
+        .map_err(ReadError)?
```
aboodman, line 163: `GetSyncHeadError`?

src/sync/mod.rs
```diff
@@ -139,6 +148,114 @@ pub enum BeginSyncError {
...
+    // Return replay commits if any.
+    if !pending.is_empty() {
+        let mut replay_mutations: Vec<Mutation> = Vec::new();
+        for c in pending {
+            let (name, args) = match c.meta().typed() {
+                MetaTyped::Local(lm) => (
+                    lm.mutator_name().to_string(),
+                    lm.mutator_args_json().to_vec(),
+                ),
+                _ => return Err(ProgrammerError("pending mutation is not local".to_string())),
```
aboodman, line 209: Can we change the name `ProgrammerError` to `InternalError` throughout? Our customers are also programmers and might not understand the error is something we did wrong, not them.

src/sync/mod.rs
```diff
@@ -139,6 +148,114 @@ pub enum BeginSyncError {
...
+        });
+    }
+
+    // TODO check invariants
+
+    // No mutations to replay so set the main head to the sync head and sync complete!
+    dag_write
+        .set_head(db::DEFAULT_HEAD_NAME, Some(&sync_head_hash))
+        .await
+        .map_err(WriteError)?;
```
aboodman, line 230: These cases should have their own error tags so we can tell them apart if/when they happen.

src/sync/mod.rs
```diff
@@ -139,6 +148,114 @@ pub enum BeginSyncError {
...
+    MissingMainHead,
+    MissingSyncHead,
+    NoBaseSnapshot(db::BaseSnapshotError),
+    OverlappingSyncs,
+    PendingError(db::PendingError),
+    ProgrammerError(String),
+    ReadError(dag::Error),
+    SyncSnapshotWithNoBasis,
+    WriteError(dag::Error),
+    WrongSyncHead,
```
aboodman, line 256: OK there are admittedly a lot of cases here but this is maybe the most complex function in the entire system and I'd argue it's useful to understand the difference between all these different cases.

aboodman:
This is really nice work @phritz .

src/db/commit.rs
```diff
@@ -210,6 +210,30 @@ impl Commit {
         Ok(commit)
     }
 
+    // Vector returned in chain-head-first order.
+    pub async fn pending(
```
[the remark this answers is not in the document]
phritz: clarified in https://github.com/rocicorp/repc/pull/193

src/db/commit.rs
```diff
@@ -210,6 +210,30 @@ impl Commit {
...
+        dag_read: &dag::Read<'_>,
+    ) -> Result<Vec<Commit>, PendingError> {
+        use PendingError::*;
+        let mut commit = Commit::from_hash(hash, dag_read)
+            .await
+            .map_err(NoSuchCommit)?;
+        let mut commits = Vec::new();
+        while !commit.meta().is_snapshot() {
+            let meta = commit.meta();
+            let basis_hash = meta
```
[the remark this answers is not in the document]
phritz: The str returned from basis_hash is borrowed from the commit's fb via meta, so we can't hold a reference to it while we move commit into the commits vector. to_string() enables the commit to be moved.

src/sync/mod.rs
```diff
@@ -18,6 +18,15 @@ use std::str::FromStr;
 
 pub const SYNC_HEAD_NAME: &str = "sync";
 
+#[derive(Debug, DeJson, SerJson)]
+pub struct Mutation {
+    id: u64,
+    name: String,
+    args: Vec<u8>,
+    #[nserde(skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
phritz: It serializes it as null: https://serde.rs/impl-serialize.html#other-special-cases. Seems like we might need to go check all of the other Options in the code and add skip_serializing_if?

src/sync/mod.rs
```diff
@@ -139,6 +148,114 @@ pub enum BeginSyncError {
...
+    }
+
+    // Ensure another sync has not landed a new snapshot on the main chain.
+    let sync_snapshot = Commit::base_snapshot(&sync_head_hash, &dag_read)
+        .await
+        .map_err(NoBaseSnapshot)?;
+    let main_head_hash = dag_read
+        .get_head(db::DEFAULT_HEAD_NAME)
+        .await
+        .map_err(ReadError)?
```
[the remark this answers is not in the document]
phritz: done in https://github.com/rocicorp/repc/pull/193

src/sync/mod.rs
phritz: done in https://github.com/rocicorp/repc/pull/193

src/sync/mod.rs
phritz: done in https://github.com/rocicorp/repc/pull/193

src/sync/mod.rs
phritz: done in https://github.com/rocicorp/repc/pull/193

src/sync/mod.rs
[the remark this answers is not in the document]
aboodman: No strong opinion.

PR #135, opened 2020-09-01
For discussion - do not merge - Explicitly rollback write transactions when closed.

aboodman:
Failing to do this before caused a JavaScript exception because the transaction would be dropped without de-registering event handlers. When the transaction closed (implicitly) the relevant events would be fired, but the corresponding Rust closures were already gone.

An alternate (and much smaller) solution would be to de-register event handlers in a custom Drop implementation, but it seems better to have all paths go through either commit() or rollback().


phritz (commented):
LGTM

PR #138, opened 2020-09-01
Finish wiring up scan().

### LINKED ISSUE #44: Scan

aboodman:
Fixes #44.

### DIFF
--- a/src/embed/types.rs
+++ b/src/embed/types.rs
@@ -1,5 +1,7 @@
 #![allow(clippy::redundant_pattern_matching)] // For derive(DeJson).
 
+use crate::db;
+use crate::prolly;
 use crate::sync;
 use crate::util::nanoserde::any;
 use nanoserde::{DeJson, SerJson};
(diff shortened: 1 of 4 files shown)


src/embed/types.rs
```diff
@@ -90,6 +92,87 @@ pub struct GetResponse {
     pub has: bool, // Second to avoid trailing comma if value == None.
 }
 
+#[derive(DeJson)]
+pub struct ScanKey {
+    value: String,
```
aboodman, line 97: We use `String` for keys and values at the bindings layer but `Vec<u8>` at key/value layer. We should figure this out someday.

src/embed/types.rs
```diff
@@ -90,6 +92,87 @@ pub struct GetResponse {
...
+    exclusive: bool,
+}
+
+impl<'a> From<&'a ScanKey> for db::ScanKey<'a> {
+    fn from(source: &'a ScanKey) -> db::ScanKey<'a> {
+        db::ScanKey {
+            value: source.value.as_bytes(),
```
aboodman, line 104: Note: doesn't allocate! whee!

src/embed/types.rs
```diff
@@ -90,6 +92,87 @@ pub struct GetResponse {
...
+#[derive(Debug)]
+pub enum FromProllyEntryError {
+    BadUTF8(std::string::FromUtf8Error),
+}
+
+impl<'a> std::convert::TryFrom<prolly::Entry<'a>> for ScanItem {
+    fn try_from(entry: prolly::Entry) -> Result<Self, Self::Error> {
+        use FromProllyEntryError::*;
+        Ok(Self {
+            key: String::from_utf8(entry.key.to_vec()).map_err(BadUTF8)?,
```
aboodman, line 164: Note: does allocate :(. Boo.

src/embed/types.rs
```diff
@@ -90,6 +92,87 @@ pub struct GetResponse {
     pub has: bool, // Second to avoid trailing comma if value == None.
 }
 
+#[derive(DeJson)]
```
phritz, line 95: I think i would have expected to see some dispatch tests for this in https://github.com/rocicorp/repc/blob/master/tests/wasm.rs. Unless there is a technical reason why we can't have them seems like in addition to the scan unit tests in db we probably want some coverage at the higher level as well.

src/embed/types.rs
phritz: heh

phritz (commented):
LGTM

src/embed/types.rs
aboodman: You got it.

Issue #151, opened 2020-09-03
Treat transaction args as opaque data


aboodman:
Currently transaction arguments are decoded in both repc and replicache-sdk-js in various locations. This is both expensive and also prevents us from easily doing useful sanity checks.

aboodman:
See https://github.com/rocicorp/repc/pull/153/commits/9e73ae592d2980a7c9f360fc1d9625b84b891564 for previous sanity check that should be re-instated with this bug.

aboodman:
See #139

phritz:
args in opentransaciton become a string https://github.com/rocicorp/repc/blob/a5ee7e5caeebdd41205046356853b22dfd0aed56/src/embed/types.rs#L13

phritz:
this forces one of three choices:
1. move to something not json at the boundaries with customers (prolly not tenable rn)
2. base64 encode values
3. or use strings

phritz:
heh:
- opentransaction args becomes utf8 string, gets stored opaquely
- we still probably need Any to encode the string when calling batchpush

PR #153, opened 2020-09-03
Remove inconsistent args

aboodman:


phritz (commented):
in the future do you imagine we dont check the args at all? if we wont we should just remove the args param. if we will we should have a TODO here referencing https://github.com/rocicorp/repc/issues/151

phritz (commented):
LGTM

PR #154, opened 2020-09-03
Implement del RPC

aboodman:
Here's ya dang rpc.

Manually tested the delete button in the todo sample works now!

Fixes #154

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -63,6 +63,7 @@ async fn connection_future<'a, 'b>(
         "get" => execute_in_txn(do_get, txns, req).await,
         "scan" => execute_in_txn(do_scan, txns, req).await,
         "put" => execute_in_txn(do_put, txns, req).await,
+        "del" => execute_in_txn(do_del, txns, req).await,
         "openTransaction" => execute(do_open_transaction, store, txns, req).await,
         "commitTransaction" => execute(do_commit, store, txns, req).await,
         "closeTransaction" => execute(do_abort, store, txns, req).await,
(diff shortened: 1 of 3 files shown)


src/embed/connection.rs
```diff
@@ -398,6 +399,17 @@ async fn do_put(txn: &RwLock<Transaction<'_>>, req: PutRequest) -> Result<PutRes
     Ok(PutResponse {})
 }
 
+async fn do_del(txn: &RwLock<Transaction<'_>>, req: DelRequest) -> Result<DelResponse, String> {
+    let mut guard = txn.write().await;
+    let write = match &mut *guard {
+        Transaction::Write(w) => Ok(w),
+        Transaction::Read(_) => Err("Specified transaction is read-only".to_string()),
+    }?;
+    let had = write.as_read().has(req.key.as_bytes());
```
aboodman, line 408: Sadly using the return value from the underlying data structure is not so simple as I made it seem.

phritz: yeah i started thinking it was going to be easy and then gave up until i had more time

src/embed/types.rs
```diff
@@ -184,6 +184,18 @@ pub struct PutRequest {
...
 pub struct PutResponse {}
 
+#[derive(DeJson)]
+pub struct DelRequest {
+    #[nserde(rename = "transactionId")]
+    pub transaction_id: u32,
+    pub key: String,
+}
+
+#[derive(DeJson, SerJson)]
```
phritz, line 194: "ok" makes it sound like if ok is false there was a problem. how about 'was_present' or 'had' similar? at the very least we should have a comment.

phritz (approved):

src/embed/types.rs
aboodman: Yeah it needs to be `ok` for compat with the js, but I can use rename so at least the field is reasonable.

aboodman: Fixed

PR #160, opened 2020-09-04
Fix bug where start key was ignored in scan from js.

aboodman:


phritz (commented):
yowch, is it scary that this wasn't caught by a test?

LGTM

aboodman:
Not really. The test that should have caught it was the js sdk's integration test. But that test wasn't up and running yet at the time this code was written. We'd have caught it later ... like a few days ago, when @arv landed the conversion of that test suite over to wasm.

Issue #159, opened 2020-09-04
Implement remaining misc RPCs


aboodman:
Thinking specifically of like dropping a database, maybe there are others.

aboodman:
- [ ] list
- [ ] drop
- [ ] version
- [ ] setLogLevel (probably ignore for now?)

PR #164, opened 2020-09-04
Implement the drop RPC.

aboodman:
Toward https://github.com/rocicorp/repc/issues/159

### DIFF
--- a/src/embed/dispatch.rs
+++ b/src/embed/dispatch.rs
@@ -54,6 +54,7 @@ async fn dispatch_loop(rx: Receiver<Request>) {
         let response = match req.rpc.as_str() {
             "open" => Some(do_open(&mut conns, &req).await),
             "close" => Some(do_close(&mut conns, &req).await),
+            "drop" => Some(do_drop(&mut conns, &req).await),
             "debug" => Some(do_debug(&conns, &req).await),
             _ => None,
         };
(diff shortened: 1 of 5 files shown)


arv (approved):

src/kv/idbstore.rs
```diff
@@ -116,6 +117,23 @@ impl IdbStore {
...
 
+    pub async fn drop_store(name: &str) -> Result<()> {
+        let window = web_sys::window().ok_or(StoreError::Str(str!("could not get window")))?;
+        let factory = window
+            .indexed_db()?
+            .ok_or(StoreError::Str(str!("could not get indexeddb")))?;
+        let request = factory.delete_database(name)?;
+        let (callback, receiver) = IdbStore::oneshot_callback();
+        request.set_onsuccess(Some(callback.as_ref().unchecked_ref()));
+        request.set_onerror(Some(callback.as_ref().unchecked_ref()));
```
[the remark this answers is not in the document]
aboodman: Yeah this is a little sketchy in that it relies on idb only calling one of the two callbacks, but it's what we do elsewhere in this file, and seems to work for now.

src/embed/dispatch.rs
```diff
@@ -131,6 +132,19 @@ async fn do_close(conns: &mut ConnMap, req: &Request) -> Response {
     Ok("".into())
 }
 
+async fn do_drop(_: &mut ConnMap, req: &Request) -> Response {
+    match &req.db_name[..] {
+        #[cfg(not(target_arch = "wasm32"))]
+        "mem" => Ok("".into()),
```
phritz, line 138: curious why just "mem"? is there some guarantee that there is exactly one db and it is named "mem" when in not-browser mode? think i might have expected _ => Ok for not wasm32 and the _ you have below for wasm32.

phritz (commented):
LGTM

src/embed/dispatch.rs
aboodman: The way that you trigger getting access to the mem db is by using the special name "mem". And that is only available in non-wasm (which is a hacky way of saying "test" mode).

PR #169, opened 2020-09-10
Implement client_id generation.

### LINKED ISSUE #45: Implement clientID

aboodman:
See: https://www.kapwing.com/e/5f5a9e88bfffe500af5e8b46 for more information.

Fixes #45

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -5,7 +5,7 @@ use crate::db;
 use crate::fetch;
 use crate::sync;
 use crate::util::nanoserde::any;
-use async_fn::{AsyncFn2, AsyncFn3};
+use async_fn::AsyncFn2;
 use async_std::stream::StreamExt;
 use async_std::sync::{Receiver, RecvError, RwLock};
 use futures::stream::futures_unordered::FuturesUnordered;
@@ -49,8 +49,7 @@ enum UnorderedResult {
 
 async fn connection_future<'a, 'b>(
     rx: &Receiver<Request>,
-    store: &'a dag::Store,
-    txns: &'b TxnMap<'a>,
+    ctx: Context<'a, 'b>,
     request: Option<Request>,
 ) -> UnorderedResult {
     let req = match request {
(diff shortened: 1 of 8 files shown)


src/embed/connection.rs
```diff
@@ -82,7 +81,7 @@ async fn connection_future<'a, 'b>(
     UnorderedResult::None()
 }
 
-pub async fn process(store: dag::Store, rx: Receiver<Request>) {
+pub async fn process(store: dag::Store, rx: Receiver<Request>, client_id: String) {
```
arv, line 84: I think I need to reread the `String` vs `&str` part but are we not copying the string too much here?

arv (approved):

src/embed/connection.rs
aboodman: `String` is a move type so what's happening here is it's moving into `process`, not getting copied. You can verify this by looking for the `Copy` trait here:

https://doc.rust-lang.org/std/string/struct.String.html#trait-implementations

By default things in Rust are move unless `Copy` is implemented.

aboodman: (the places you see `.clone()` we are explicitly copying the string, this is sort of unavoidable given current shape of things.

src/embed/dispatch.rs
```diff
@@ -151,3 +156,30 @@ async fn do_debug(conns: &ConnMap, req: &Request) -> Response {
...
     }
 }
+
+#[derive(Debug)]
+enum InitClientIdError {
+    GetErr(StoreError),
+    OpenErr(StoreError),
+    InvalidUtf8(std::string::FromUtf8Error),
+    PutClientIdErr(StoreError),
+    CommitErr(StoreError),
```
phritz, line 166: alpha mo betta

src/embed/dispatch.rs
```diff
@@ -151,3 +156,30 @@ async fn do_debug(conns: &ConnMap, req: &Request) -> Response {
...
+}
+
+async fn init_client_id(s: &dyn Store) -> Result<String, InitClientIdError> {
```
phritz, line 169: i prefer get_or_init_client_id

src/util/uuid/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+#[cfg(target_arch = "wasm32")]
+fn make_random_numbers(numbers: &mut [u8]) {
+    web_sys::window()
+        .expect("window is not available")
+        .crypto()
+        .expect("window.crypto is not available")
+        .get_random_values_with_u8_array(numbers)
+        .expect("window.crypto.getRandomValues not available");
```
phritz, line 18: better to return an error?

src/util/uuid/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+        .iter()
+        .enumerate()
+        .map(|(i, kind)| match kind {
+            UuidElements::Random09AF => {
+                char::from_digit((random_numbers[i] & 0b1111) as u32, 16).expect(ERROR_MAKE_CHAR)
+            }
+            UuidElements::Random89AB => {
+                char::from_digit((random_numbers[i] & 0b11) as u32 + 8, 16).expect(ERROR_MAKE_CHAR)
+            }
+            UuidElements::Version => '4',
```
phritz, line 89: curious where this version comes from?

phritz (commented):
LGTM!

src/embed/dispatch.rs
```diff
@@ -151,3 +156,30 @@ async fn do_debug(conns: &ConnMap, req: &Request) -> Response {
...
+    use InitClientIdError::*;
+
+    const CID_KEY: &str = "sys/cid";
+    let cid = s.get(CID_KEY).await.map_err(GetErr)?;
+    if let Some(cid) = cid {
+        let s = String::from_utf8(cid).map_err(InvalidUtf8)?;
+        return Ok(s);
+    }
+    let wt = s.write().await.map_err(OpenErr)?;
```
phritz, line 178: should probably test that it returns the same client id on a second call

src/util/uuid/mod.rs
aboodman: It's part of the spec. @arv being @arv, he implemented the actual uuid spec rather than a random number which is all we needed :).

aboodman: In particular v4 of the uuid spec. Yes there have been 4 (actually 5) versions of a spec for a large unique number.

arv: This was taken from somewhere... I think the original PR had a link to it.

If we only need a random number we cam simplify.

arv:
Please see https://github.com/rocicorp/repc/pull/59#issuecomment-662754922 I feel like these things were not fully taken care of. Especially the license and the code size

aboodman:
Thanks, I have a backlog of code review stuff to circle back on.

src/embed/dispatch.rs
aboodman: Fixing in upcoming review.

src/embed/dispatch.rs
aboodman: Rejected.

src/util/uuid/mod.rs
aboodman: In principle yes, but changing this turned into a big yak shave in embed, so I'm putting it off for now.

src/util/uuid/mod.rs
aboodman: How big does this really need to be @phritz . IIRC clientID is scoped within a particular account. Two random 64b numbers should more than do it, right? Should we go ahead and simplify this?

aboodman: https://github.com/rocicorp/repc/issues/183

src/embed/dispatch.rs
aboodman: Done in next pr.

Issue #173, opened 2020-09-14
"close" does not close the IDB db


arv:
We need to call IDBDatabase.close when closing the replicache database.

arv:
Fixed in #174**

PR #174, opened 2020-09-15
Make sure we call close on the IDBDatabase

arv:
We were not closing the IDB database when we closed the Replicache db

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -69,7 +69,13 @@ async fn connection_future<'a, 'b>(
         "beginSync" => execute(ctx, do_begin_sync, req).await,
         "maybeEndSync" => execute(ctx, do_maybe_end_sync, req).await,
         "close" => {
-            req.response.send(Ok("".into())).await;
+            let result = ctx
+                .store
+                .close()
+                .await
+                .map(|_| "".into())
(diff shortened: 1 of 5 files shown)


arv:
I verified that `drop` works after this. So, this explains why drop was not working correctly before.

src/embed/connection.rs
```diff
@@ -69,7 +69,13 @@ async fn connection_future<'a, 'b>(
         "beginSync" => execute(ctx, do_begin_sync, req).await,
         "maybeEndSync" => execute(ctx, do_maybe_end_sync, req).await,
         "close" => {
-            req.response.send(Ok("".into())).await;
+            let result = ctx
```
aboodman, line 72: I think you may want to swap the order of these two calls. You are closing the idb store *then* stopping the message loop that is using it. It seems like this could result in a few stray calls still getting through and running on the closed store.

aboodman (approved):

arv:
Fixes #173

src/embed/connection.rs
arv: Which two calls? Are you saying `send` the result before calling `close`?

aboodman: I'm an idiot, nevermind.

Issue #161, opened 2020-09-04
We need to correctly implement onversionchange


aboodman:
See: https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB#Version_changes_while_a_web_app_is_open_in_another_tab and https://twitter.com/aboodman/status/1301969990904475648.

I thought that we don't need to implement the version change dance of IDB because we don't ever change our IDB schema. *However* we do expose dropping the database through the API which counts as a version change. Also the user can drop the DB through Chrome's dev tools UI which might also count (should check).

Luckily we abstract all the away from our user, so we can just implement `onversionchange` to close and re-open the db.

arv:
We talked a bit about this and our understanding is that we need to close the IDBDatabase to "confirm" drops happening out of bounds.

This close needs to be routed back to the js sdk. It can be done with a callback or by returning a specific error when incove is called. Then JS can cleanup its state (subscriptions etc).

aboodman:
Is this done @arv?

arv:
I think there was a reason why I marked it as towards instead of closes... but I cannot recall. I think we talked about routing back the close to the Replicache object but at the moment all the dispatch calls will fail and we do not really distinguish the different flavors of errors.

I think I wanted to deal with db is closed errors in the bindings and that was why I kept this open.

PR #191, opened 2020-09-21
Close the db when onversionchange

arv:
We now close the database if it is removed.

We do not yet update the JS Replicache object that the database got
closed. We can either:
 - Later calls to the database throws and we can then update the js
   state.
 - Add a callback

Towards #161

### DIFF
--- a/src/kv/idbstore.rs
+++ b/src/kv/idbstore.rs
@@ -94,18 +94,10 @@ impl IdbStore {
         };
         let request = factory.open(name)?;
         let (callback, receiver) = IdbStore::oneshot_callback(logger.clone());
-        let request_copy = request.clone();
         let closure_logger = logger.clone();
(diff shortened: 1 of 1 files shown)


src/kv/idbstore.rs
```diff
@@ -94,18 +94,10 @@ impl IdbStore {
...
-                Ok(r) => r,
-                Err(e) => {
-                    closure_logger.error(format!("Error before ugradeneeded: {:?}", e));
-                    return;
-                }
-            };
-            let db = web_sys::IdbDatabase::unchecked_from_js(result);
-
+        let onupgradeneeded = Closure::once(move |event: web_sys::IdbVersionChangeEvent| {
+            let target = event.target().unwrap();
```
aboodman, line 99: Don't panic. But here in handling some random event, we can't return an error code. So `error!` maybe?

arv: I should not have changed this function. I'm reverting that

PR #192, opened 2020-09-22
For discussion - deloop dispatch

aboodman:
This PR doesn't work but shows my train of thought.

We change `dispatch` into a "normal" function that delegates directly to the "doer" functions. I already tested (see `do_test`) that the basic concept works -- the futures get executed and calls can overlap this way.

However, I rediscovered why we need the dispatch loop!

It's fundamentally because of the relationship between transactions and stores. Transactions have a reference to their store, so the store must outlive the transaction. In particular the `RWLockGuard` in `kv::{Read,Write}Transaction`.

My thought was that the RWLock would let us get around this, but it does not. What it weakens is the single writer requirement. The reference requirement remains.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4785bdd1c96b2a846b2bd7cc02e86b6b3dbf14e7e53446c4f54c92a361040822"
 
 [[package]]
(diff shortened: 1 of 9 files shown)


phritz (commented):
Thanks for sharing, not surprised there are lifetime issues. I'm not going to spend time grokking this fully tho unless you want me to.

PR #193, opened 2020-09-22
address a bunch of code review comments

phritz:
separate logical commits

### DIFF
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -242,9 +242,9 @@ impl Commit {
     }
 
     // Parts are (last_mutation_id, server_state_id).
-    pub fn snapshot_meta_parts(c: &Commit) -> Result<(u64, String), ProgrammerError> {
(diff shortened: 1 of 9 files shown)


src/db/commit.rs
```diff
@@ -241,7 +241,9 @@ impl Commit {
     // Parts are (last_mutation_id, server_state_id).
     pub fn snapshot_meta_parts(c: &Commit) -> Result<(u64, String), InternalProgrammerError> {
         match c.meta().typed() {
-            MetaTyped::Local(_) => Err(InternalProgrammerError::WrongType(str!("Snapshot meta expected"))),
+            MetaTyped::Local(_) => Err(InternalProgrammerError::WrongType(str!(
```
aboodman, line 244: not worth going back for, but i'd also prefer to just stick to the convention of one enum per method for a few reasons:

- it's nice to have a convention so that readers don't have to think too much
- i still hope to someday mechanize the convention

phritz: yeah i'm board with that now. this was just a mechanical rename carried out by the editor i dont even think i saw the line :)

PR #200, opened 2020-09-24
For discussion - Serialize to unsafe view

aboodman:
Only the last commit is interesting. I had originally thought I'd return batches of ArrayBuffers back to client but cannot because that would cause allocations after Uint8Array::view() which can invalidate them.


aboodman:
This increases scan performance another 76%, to 30MB/s (with incompressible data). can't be landed as-is, have to find some reasonable way to pass the js handler to rust.

PR #202, opened 2020-09-28
Make basis and original hash references strong

### LINKED ISSUE #105: Commit: Need to make the basis hash reference "strong"
Currently the commit struct knows its basis, but it's a weak reference in the sense that we don't tell dag about the relationship:

https://github.com/rocicorp/repc/blob/master/src/db/commit.rs#L188

This means that if we had GC, these basis pointers would not keep the referenced commits alive. Before we implement GC, we need to make these pointers strong.

arv:
Fixes #105

### DIFF
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -7,6 +7,28 @@ use str_macro::str;
 
 pub const DEFAULT_HEAD_NAME: &str = "main";
 
+#[derive(Clone, Copy)]
+enum Ref<'a> {
+    Strong(&'a str),
+    Weak(&'a str),
+}
+
+impl<'a> Ref<'a> {
+    pub fn hash(self) -> &'a str {
+        match self {
+            Ref::Weak(s) => s,
+            Ref::Strong(s) => s,
+        }
+    }
+
+    pub fn strong_or_none(self) -> Option<&'a str> {
+        match self {
+            Ref::Weak(_) => None,
+            Ref::Strong(s) => Some(s),
+        }
+    }
+}
+
 // Commit is a thin wrapper around the Commit flatbuffer that makes it
 // easier to read and write them. Commit::load() does validation
 // so that users don't have to worry about missing fields.
@@ -39,11 +61,12 @@ impl Commit {
         Commit::new_impl(
             builder,
             local_create_date,
-            basis_hash,
+            basis_hash.map(Ref::Strong),
             checksum,
             commit_fb::MetaTyped::LocalMeta,
             local_meta.as_union_value(),
-            value_hash,
+            Ref::Strong(value_hash),
+            original_hash.map(Ref::Strong),
         )
     }
 
@@ -64,11 +87,12 @@ impl Commit {
         Commit::new_impl(
             builder,
             local_create_date,
-            basis_hash,
+            basis_hash.map(Ref::Weak),
             checksum,
             commit_fb::MetaTyped::SnapshotMeta,
             snapshot_meta.as_union_value(),
-            value_hash,
+            Ref::Strong(value_hash),
+            None,
         )
     }
 
@@ -162,15 +186,16 @@ impl Commit {
     fn new_impl(
         mut builder: FlatBufferBuilder,
         local_create_date: &str,
-        basis_hash: Option<&str>,
+        basis_hash: Option<Ref>,
         checksum: Checksum,
         union_type: commit_fb::MetaTyped,
         union_value: flatbuffers::WIPOffset<flatbuffers::UnionWIPOffset>,
-        value_hash: &str,
+        value_hash: Ref,
+        original_hash: Option<Ref>,
     ) -> Commit {
         let meta_args = &commit_fb::MetaArgs {
             local_create_date: builder.create_string(local_create_date).into(),
-            basis_hash: basis_hash.map(|s| builder.create_string(s)),
+            basis_hash: basis_hash.map(|r| builder.create_string(r.hash())),
             checksum: builder.create_string(&checksum.to_string()).into(),
             typed_type: union_type,
             typed: union_value.into(),
@@ -178,12 +203,18 @@ impl Commit {
         let meta = commit_fb::Meta::create(&mut builder, meta_args);
         let commit_args = &commit_fb::CommitArgs {
             meta: meta.into(),
-            value_hash: builder.create_string(value_hash).into(),
+            value_hash: builder.create_string(value_hash.hash()).into(),
         };
         let commit = commit_fb::Commit::create(&mut builder, commit_args);
         builder.finish(commit, None);
 
-        let chunk = dag::Chunk::new(builder.collapse(), &[value_hash]);
+        let refs = std::iter::once(value_hash)
+            .chain(basis_hash)
+            .chain(original_hash)
+            .filter_map(Ref::strong_or_none)
+            .collect::<Vec<&str>>();
+
+        let chunk = dag::Chunk::new(builder.collapse(), &refs);
         Commit { chunk }
     }
 
@@ -495,27 +526,34 @@ mod tests {
             let actual = Commit::from_chunk(chunk);
             assert_eq!(expected, actual);
         }
-        test(
-            make_commit(
-                Some(Box::new(|b: &mut FlatBufferBuilder| {
-                    make_local_meta(b, 0, "".into(), Some(&[]), "".into())
-                })),
-                "".into(),
-                "".into(),
-                checksum_str,
-                "".into(),
-            ),
-            Ok(Commit::new_local(
-                "",
-                "".into(),
-                checksum,
-                0,
-                "",
-                &[],
-                "".into(),
-                "",
-            )),
-        );
+        for basis_hash in &[None, Some(""), Some("hash")] {
+            test(
+                make_commit(
+                    Some(Box::new(|b: &mut FlatBufferBuilder| {
+                        make_local_meta(b, 0, "".into(), Some(&[]), "original".into())
+                    })),
+                    "".into(),
+                    *basis_hash,
+                    checksum_str,
+                    "value".into(),
+                    &(if basis_hash.is_none() {
+                        vec!["value", "original"]
+                    } else {
+                        vec!["value", basis_hash.unwrap(), "original"]
+                    }),
+                ),
+                Ok(Commit::new_local(
+                    "",
+                    *basis_hash,
+                    checksum,
+                    0,
+                    "",
+                    &[],
+                    "original".into(),
+                    "value",
+                )),
+            );
+        }
         test(
             make_commit(
                 Some(Box::new(|b: &mut FlatBufferBuilder| {
@@ -525,6 +563,7 @@ mod tests {
                 "".into(),
                 checksum_str,
                 "".into(),
+                &["", ""],
             ),
             Err(LoadError::MissingMutatorName),
         );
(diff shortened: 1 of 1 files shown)


src/db/commit.rs
```diff
@@ -183,7 +183,11 @@ impl Commit {
         let commit = commit_fb::Commit::create(&mut builder, commit_args);
         builder.finish(commit, None);
 
-        let chunk = dag::Chunk::new(builder.collapse(), &[value_hash]);
+        let refs = std::iter::once(value_hash)
```
aboodman, line 186: Fancy. Another way to say it would just be:

```
let mut refs = vec![value_hash];
if let Some(basis_hash) = basis_hash {
  refs.push_back(basis_hash);
}
```

arv: I had this before and switch to the `once` to keep the two places more similar

aboodman:
This brings up the next point that of course we don't want all commits to strongly reference their basis. Because we want to actually delete history. But the local pending ones do need to ... because we need it during sync.

@phritz what should the actual policy be?

aboodman:
Is it a true statement that for sync, we only need pending commits to strongly retain their basis?

arv:
That seems right to me but I would like to get that confirmed.

phritz:
> Is it a true statement that for sync, we only need pending commits to strongly retain their basis?

I think it's true to say that we need:
- all the local commits on the main chain to strongly retain their basis (this will include the base snapshot when pending commits exist)
- all heads to strongly retain the thingy they point to (this will include the base snapshot when pending commits do not exist)
- all local commits on the sync chain to strongly retain their basis (so we retain the sync chain)

Said another way, we need to retain all commits from the main and sync heads to the base snapshot.

We have to decide what to do about two cases:
1. Replayed commits: when we switch the main head to the sync head to complete sync, replayed commits will have a reference to local commits via `original` that no longer retained by the above rules. These original commits were previously on the main chain but no head is pointing to them anymore.
2. The basis of the base snapshot: the base snapshot is retained under the above rules but when we finalize sync we switch the main head to point to the sync head. The sync snapshot becomes the base snapshot and we need to decide what happens to the previous sync snapshot, the current sync snapshot's basis?

phritz:
I guess the question is how aggressively we want to prune old commits, specifically precious base snapshots we have synced on top of and old replay commits that we have replayed on a new snapshot. My sense is that we should not keep these old commits around.

aboodman:
On Mon, Sep 28, 2020 at 12:07 PM Phritz <notifications@github.com> wrote:

> I guess the question is how aggressively we want to prune old commits,
> specifically precious base snapshots we have synced on top of and old
> replay commits that we have replayed on a new snapshot. My sense is that we
> should not keep these old commits around.
>
I get the sense it will be easier to reason about if we remove everything
that isn't strictly necessary. If you agree, let's do that.

>

src/db/commit.rs
```diff
@@ -183,7 +183,11 @@ impl Commit {
...
+            .chain(basis_hash.into_iter())
```
phritz, line 187: we need to decide whether to include localmeta.original_hash in the refs

phritz:
If we remove everything that's not strictly necessary then we have to rewrite the sync chain in order to finalize sync. We need to update the sync snapshot to remove its basis and then each replayed commit on top has to have its basis pointed to its new parent. (If we make original strong, we also have to reset original in each of the replayed commits as we go. If we don't make it strong, it can just dangle. But preference for making it strong and cleaning it up.)

src/db/commit.rs
arv: I'm trying to page all of this back into memory... Do we need to be able to read data from the original hash?

arv:
The goal to remove everything that is not strictly necessary seems like the right thing to do. We do not want to use up more space than we need.

arv:
We talked about this on VC and I did some more researching of the existing code.

Goals:
- Only keep required data
- Minimize fixups/rewrites

`original_hash` is currently read to compare the `name` so we need to keep it alive as long as we have the local commit. When the local commit goes away (replaced by a snapshot commit) the original local commit will also go away.

We have "GC roots" for `head` and the `sync_head`.

The only thing keeping a snapshot commit alive is a local commit, the `head` or the `sync_head`.

Things that keep a local commit alive is `head`, `sync_head` or it being the `basis` or `original` of another local commit.

Once we are fully synced there will be a single snapshot commit, referenced by `head`. Everything else can be collected.

This has no fixups. Things will go away as needed and once there is only a snapshot commit it will have a dangling reference to the `basis` which has now been collected.

phritz:
Not sure what "no fixups" means, we need do to address the two things that I pointed out: 
- the sync snapshot currently has the base snapshot as its basis, so we either need to not set the sync snapshot basis or remove it when when finalizing sync and update the replayed commit bases (because removing a commit's basis changes its hash, and therefor its children need to be updated all the way to the head). i have some preference for keeping the sync snapshot basis as the base snapshot so that we can check invariants between the two just before we finalize sync, but we could do that in beginsync instead if we wanted -- just less of a tight guarantee.
- replayed local commits on the sync chain have (presumably strong) original_hash pointers to pending commits. When sync is finalized we want the originals to go away. In order to make that happen we have to reset the original_hash of each replayed commit to not point to them. This necessarily changes the replayed commits' hashes and so we have to rewrite them.

phritz:
Possibly one bit of missing info is that when we are done syncing head does not always point to the new base snapshot. Sometimes there are also local replayed commits on top of it. When the sync head becomes the new main head it can point to one of these local commits, which are still pending.

aboodman:
I think I am in favor of making snapshot basis pointers always weak for simplicity because it avoids “fixing up” the commit chain after sync.

However I do see it weakens some invariants. Sync has to rely on the fact that the main head is keeping sync snapshots basis alive. I think this is correct trade off but fritz is closest to this code so his preference should override mine.

I think the only thing I’m firm on is that we shouldn’t ever retain any data not needed for sync.

phritz:
Plan LGTM. Let me know when to review this PR.

arv:
Maybe I'm missing something but the concept of weak vs strong is just an item in the refs entry. It does not change the chunk data. It is OK to have a string, with a hash that used to point to a chunk that is now removed.

We can therefore change `original` from strong to weak as we need without changing the data or the hash and we do not to rewrite (fixup) any chunks 

> Not sure what "no fixups" means

I mean that we do not want to rewrite chunks unless we have to.

> Possibly one bit of missing info is that when we are done syncing head does not always point to the new base snapshot. Sometimes there are also local replayed commits on top of it. When the sync head becomes the new main head it can point to one of these local commits, which are still pending.

That is fine; the local commits keep the snapshot commit alive.

arv:
PTAL

Maybe the `enum Ref` is a bit of overkill? I learned a lot by trying that approach though.

phritz:
> Maybe I'm missing something but the concept of weak vs strong is just an item in the refs entry. It does not change the chunk data. It is OK to have a string, with a hash that used to point to a chunk that is now removed.

Right. IF we wanted to remove the ref for cleanliness to not have dangling pointers THEN it would change the hash. But if we are comfortable with them dangling then no worries.

src/db/commit.rs
```diff
@@ -7,6 +7,40 @@ use str_macro::str;
 
 pub const DEFAULT_HEAD_NAME: &str = "main";
 
+#[derive(Clone, Copy)]
+enum Ref<'a> {
+    None,
```
phritz, line 12: Might be slightly more rusty be to have the enum have Strong and Weak and the field that holds it be an Option<Ref>.

src/db/commit.rs
```diff
@@ -162,15 +198,16 @@ impl Commit {
     fn new_impl(
         mut builder: FlatBufferBuilder,
         local_create_date: &str,
-        basis_hash: Option<&str>,
+        basis_hash: Ref,
         checksum: Checksum,
         union_type: commit_fb::MetaTyped,
         union_value: flatbuffers::WIPOffset<flatbuffers::UnionWIPOffset>,
         value_hash: &str,
+        original_hash: Option<&str>,
```
phritz, line 206: the original is in the meta if you want to pull it out of there instead of passing it in, which would be more DRY.

src/db/commit.rs
```diff
@@ -183,7 +220,15 @@ impl Commit {
         let commit = commit_fb::Commit::create(&mut builder, commit_args);
         builder.finish(commit, None);
 
-        let chunk = dag::Chunk::new(builder.collapse(), &[value_hash]);
+        let mut refs = vec![value_hash];
+        if let Ref::Strong(basis_hash) = basis_hash {
```
phritz, line 224: It was my understanding that basis was only going to be strong for local commits?

src/db/commit.rs
```diff
@@ -183,7 +220,15 @@ impl Commit {
...
+            refs.push(basis_hash);
+        }
+        if let Some(original_hash) = original_hash {
```
phritz, line 227: I think we have to change this to a weak reference when sync is complete and remove the original from refs. But now that I think about it, that would actually change the hash. If original were a str instead of a typed thing then 'making it a weak ref' is just removing it from the refs -- the commit is untouched.

phritz (commented):
What's the mechanism by which heads are strong refs?

src/db/commit.rs
```diff
@@ -7,6 +7,40 @@ use str_macro::str;
...
 pub const DEFAULT_HEAD_NAME: &str = "main";
 
+#[derive(Clone, Copy)]
+enum Ref<'a> {
+    None,
+    Strong(&'a str),
+    Weak(&'a str),
+}
+
+impl<'a> Ref<'a> {
```
phritz, line 17: One more thought: it makes me a little nervous that whether something is a strong or weak ref is not tied to whether the thing is in the refs or not. Seems like we want one to imply the other.

src/db/commit.rs
arv: Yup, `new_impl` is internally used by `new_local` and `new_snapshot`

src/db/commit.rs
arv: Once again I might be mistaken but based on looking at the code...

> If original were a str instead of a typed thing

We do not have a typed thing for refs. We have a string stored in the `data` and it is up to the person creating the chunk to also include the hash in the `refs` that are strong.

Changing the refs does not change the hash.

src/db/commit.rs
arv: It is in `(union_type, union_value)`. I'll see if I can extract the data from that.

arv (commented):
Thanks

src/db/commit.rs
arv: I could not figure out how to go back to a LocalMeta here

src/db/commit.rs
arv: It did imply that but it is all inside `new_impl`

src/db/commit.rs
arv: Thanks. That did make things simpler since we can reuse all the rust building blocks

src/db/commit.rs
phritz: Gah, my bad. I overlooked that the notion of strong or weak isn't represented at the commit fb level. I guess because how we've used the dag historically I find that surprising, like if we wanted to verify that commit chains are well-formed we'd have to use our historical knowledge of the semantics of each field containing a hash at the time the commit was written, instead of ignoring weak refs and checking that all strong refs exist. But in the constrained way we are using commits in repc I guess it makes sense. 

I also forgot that chunk hash is just over the data and not data+refs. This is both a cool feature because we can delete refs under the hood when they are no longer needed and a little scary that we might inadvertently overwrite write a commit with the same data but different refs. Like if at any time in two places we want to create a strong ref -- say in order to make the named head a strong ref -- we might be tempted to use a chunk with no data but set its refs. However *all* chunks with empty data share the same hash even if they have different refs, so we could easily clobber ourselves. As another example suppose we chunked the contents in the value map we reference in each commit in different ways. Then we might end up with two chunks holding the same data and thus with the same hash, but with different refs. 

Anyway, I think you are clear on the underlying thing I was trying to say, which is that when we finalize sync for each replayed local commit we need to remove its original hash from its chunk's refs. Yes?

I will take my answer to 'how are we going to make named heads strong refs' off the air as it is not in scope for your set of changes here.

Issue #155, opened 2020-09-03
move dispatch tests 


phritz:
right now dispatch tests only run in wasm and are found separate from the dispatch code, in `/tests/`. per https://github.com/rocicorp/repc/issues/77 we should:
- move dispatch tests into `embed/` alongside their code, and have all of them run in not-wasm
- have the subset of tests that can also run in wasm by calling them from `tests/wasm.rs` (eg if they are in a specific module in embed)

PR #203, opened 2020-09-29
Cleanup connection.rs 

aboodman:
Rework connection.rs to make it a bit easier to work with.
    
This isn't an architectural rethink, just mechanical cleaning. The main thing I wanted to achieve was:
    
- reduce duplication between execute() and execute_in_txn()
- reduce coupling of "doer" functions to json inputs, so that scan can take a js function as an argument to call as a perf improvement.

No perf impact from this PR. Reduced binary size by about 5k though.

(commits can be reviewed independently)

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -157,6 +157,7 @@ impl<'a, 'b> Context<'a, 'b> {
 }
 
 #[derive(Debug)]
+#[allow(clippy::enum_variant_names)]
 enum ExecuteError {
     TransactionNotFound(u32),
     TransactionRequired,
@@ -415,13 +416,13 @@ async fn do_get_root<'a, 'b>(
     })
 }
 
-async fn do_has<'a>(txn: db::Read<'a>, req: HasRequest) -> Result<HasResponse, ()> {
+async fn do_has(txn: db::Read<'_>, req: HasRequest) -> Result<HasResponse, ()> {
     Ok(HasResponse {
         has: txn.has(req.key.as_bytes()),
     })
 }
 
-async fn do_get<'a>(read: db::Read<'a>, req: GetRequest) -> Result<GetResponse, String> {
+async fn do_get(read: db::Read<'_>, req: GetRequest) -> Result<GetResponse, String> {
     #[cfg(not(default))] // Not enabled in production.
     if req.key.starts_with("sleep") {
         use async_std::task::sleep;
@@ -448,7 +449,7 @@ async fn do_get<'a>(read: db::Read<'a>, req: GetRequest) -> Result<GetResponse, 
     })
 }
 
-async fn do_scan<'a>(read: db::Read<'a>, req: ScanRequest) -> Result<ScanResponse, String> {
+async fn do_scan(read: db::Read<'_>, req: ScanRequest) -> Result<ScanResponse, String> {
     use std::convert::TryFrom;
     let mut res = Vec::<ScanItem>::new();
     for pe in read.scan((&req.opts).into()) {
@@ -457,12 +458,12 @@ async fn do_scan<'a>(read: db::Read<'a>, req: ScanRequest) -> Result<ScanRespons
     Ok(ScanResponse { items: res })
 }
 
-async fn do_put<'a>(write: &mut db::Write<'a>, req: PutRequest) -> Result<PutResponse, ()> {
+async fn do_put(write: &mut db::Write<'_>, req: PutRequest) -> Result<PutResponse, ()> {
     write.put(req.key.as_bytes().to_vec(), req.value.into_bytes());
     Ok(PutResponse {})
 }
 
-async fn do_del<'a>(write: &mut db::Write<'a>, req: DelRequest) -> Result<DelResponse, ()> {
+async fn do_del(write: &mut db::Write<'_>, req: DelRequest) -> Result<DelResponse, ()> {
     let had = write.as_read().has(req.key.as_bytes());
     write.del(req.key.as_bytes().to_vec());
     Ok(DelResponse { had })
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -131,7 +131,7 @@ where
     S: serde::Serialize,
     F: for<'r, 's> AsyncFn3<&'r RwLock<Transaction<'s>>, T, LogContext, Output = Result<S, String>>,
 {
-    let request: T = match deserialize(&req.data) {
+    let request: T = match deserialize(&req.data.as_string().unwrap()) {
         Ok(v) => v,
         Err(e) => return req.response.send(Err(e)).await,
     };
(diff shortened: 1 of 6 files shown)


src/embed/connection.rs
```diff
@@ -186,30 +156,62 @@ impl<'a, 'b> Context<'a, 'b> {
...
+        "commitTransaction" => return to_json(do_commit(ctx, from_json(&req.data)?).await),
+        "closeTransaction" => return to_json(do_close(ctx, from_json(&req.data)?).await),
+        "beginSync" => return to_json(do_begin_sync(ctx, from_json(&req.data)?).await),
+        "maybeEndSync" => return to_json(do_maybe_end_sync(ctx, from_json(&req.data)?).await),
+        _ => (),
     };
 
-    req.response.send(result).await
+    // require read txn
+    let txn_req: TransactionRequest = from_json(&req.data)?;
```
aboodman, line 181: Part of the trick I realized is to parse the input data twice -- once to get its `transaction_id` and again to get the per-rpc params. This is non-ideal, but I think worth it. Otherwise, it is very difficult to find a way to share code between the transaction-requiring and non-transaction-requiring paths.

src/embed/connection.rs
```diff
@@ -186,30 +156,62 @@ impl<'a, 'b> Context<'a, 'b> {
...
+    let txn_id = txn_req
+        .transaction_id
+        .ok_or(TransactionRequired)
+        .map_err(to_debug)?;
+    let txn_id_string = txn_id.to_string();
+    req.lc.add_context("txid", &txn_id_string);
+    let txns = ctx.txns.read().await;
```
aboodman, line 188: This is how it was before, but this bit concerns me. It looks to me like an addition to the transaction map (e.g., `openTransaction`, even a read-only transaction) requires waiting for all other transactions to close?

src/embed/dispatch.rs
```diff
@@ -24,24 +20,18 @@ pub struct Request {
...
 type Response = Result<String, String>;
 
 lazy_static! {
     static ref SENDER: Mutex<Sender::<Request>> = {
         let (tx, rx) = channel::<Request>(1);
-
-        #[cfg(target_arch = "wasm32")]
         spawn_local(dispatch_loop(rx));
-
-        #[cfg(not(target_arch = "wasm32"))]
```
phritz, line 40: Are we sure this leaves non-wasm dispatch in a usable state? Per https://github.com/rocicorp/repc/issues/155 we are *not* running dispatch tests in not-wasm (everyone keeps shrugging this off but it seems important).

src/embed/dispatch.rs
```diff
@@ -123,15 +111,11 @@ async fn do_open(conns: &mut ConnMap, req: &Request) -> Response {
         return Ok("".into());
     }
 
-    let kv: Box<dyn Store> = match &req.db_name[..] {
-        #[cfg(not(target_arch = "wasm32"))]
```
phritz, line 126: Wait, why do we not have this anymore?

src/embed/connection.rs
```diff
@@ -7,13 +7,13 @@ use crate::sync;
 use crate::util::rlog;
 use crate::util::rlog::LogContext;
 use crate::util::to_debug;
-use async_fn::{AsyncFn2, AsyncFn3};
```
phritz, line 10: HOORAY!!

src/embed/connection.rs
phritz: def worth it. didn't even occur to me.

src/embed/connection.rs
phritz: Hmmm. I'm not sure what the intent was. I think the intent might have been that txns mutex just protects the map of transaction to id, in which case anyone who acquires it should do their thing (read or insert or remove or whatever) and immediately drop the lock. There is at least one other place besides here where that is not happening.

That intent would not cover ensuring that only one dispatch call was processing a given transaction at a time, which would require I think a lock per transaction. Not sure if that is important. 

This seems like something we need to discuss more.

src/embed/connection.rs
```diff
@@ -562,26 +532,6 @@ enum CloseTransactionError {
...
-}
-
-impl_transaction_request!(HasRequest);
-impl_transaction_request!(GetRequest);
-impl_transaction_request!(ScanRequest);
-impl_transaction_request!(PutRequest);
-impl_transaction_request!(DelRequest);
-
 // Note: dispatch is mostly tested in tests/wasm.rs.
 // TODO those tests should move here and *also* be run from there so we have
```
phritz, line 537: sigh

tests/wasm.rs
```diff
@@ -290,7 +290,7 @@ async fn test_get_put_del() {
     // Check request parsing, both missing and unexpected fields.
     assert_eq!(
         dispatch(db, "put", "{}").await.unwrap_err(),
-        "InvalidJson(missing field `transactionId` at line 1 column 2)"
+        "TransactionRequired"
```
phritz, line 293: Transaction*Id*Required?

phritz (commented):
this does improve dispatch quite a bit but I'm a little unsure about how it plays out for not-wasm (see comments)

src/embed/dispatch.rs
```diff
@@ -82,7 +82,8 @@ async fn dispatch_loop(rx: Receiver<Request>) {
     }
 }
 
-pub async fn dispatch(db_name: String, rpc: String, data: String) -> Response {
+pub async fn dispatch(db_name: String, rpc: String, data: wasm_bindgen::JsValue) -> Response {
+    let data = data.as_string().unwrap();
```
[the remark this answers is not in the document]
arv: The rpc is private API. If someone uses it directly all bets are off.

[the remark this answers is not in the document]
aboodman: Agree with fritz. We have adopted a “no panics” policy in repc because we’re embedded in someone else’s memory space. Rust makes it easy and elegant to exhaustively handle error cases. This was just an overlook.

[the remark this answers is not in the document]
aboodman: Note @phritz killing "the application" here is a somewhat squishy concept. Wasm is a VM - when it dies the app doesn't completely die ... like it doesn't sadtab or segfault. Just the Wasm VM crashes.

In practice I'm sure this makes most apps unusable though and it's not something that we want devs to have to deal with, so this is just a pedantic point.

src/embed/connection.rs
aboodman: https://github.com/rocicorp/repc/issues/240

src/lib.rs
```diff
@@ -1,8 +1,8 @@
 #[macro_use]
 pub mod util;
 
-#[cfg(not(target_arch = "wasm32"))]
-mod ffi;
+//#[cfg(not(target_arch = "wasm32"))]
```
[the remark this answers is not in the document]
aboodman: It was intentional, as we discussed the other day. Not going to address this now that I think you've taken it over.

tests/wasm.rs
aboodman: https://github.com/rocicorp/repc/pull/247

PR #205, opened 2020-09-29
Scan: pass each scan item directly to JS through an ArrayBuffer as soon as we have it.

aboodman:
This improves scan perf by another ~30%.

replicache_client.js: 24484
replicache_client.js.br: 4601
replicache_client_bg.wasm: 516997
replicache_client_bg.wasm.br: 149987

Running benchmarks please wait...
populate 1024x1000 (clean) x 2.22 MB/s ±0.0% (0 runs sampled)
populate 1024x1000 (dirty) x 1.66 MB/s ±0.0% (0 runs sampled)
scan 1024x1000 x 25.70 MB/s ±0.0% (0 runs sampled)
scan 1024x5000 x 29.41 MB/s ±0.0% (0 runs sampled)

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -10,10 +10,11 @@ use crate::util::to_debug;
 use async_std::stream::StreamExt;
 use async_std::sync::{Receiver, RecvError, RwLock};
 use futures::stream::futures_unordered::FuturesUnordered;
+use js_sys::{Function, Reflect, Uint8Array};
 use log::warn;
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicU32, Ordering};
-use wasm_bindgen::JsValue;
+use wasm_bindgen::{JsCast, JsValue};
 
 lazy_static! {
     static ref TRANSACTION_COUNTER: AtomicU32 = AtomicU32::new(1);
@@ -207,7 +208,9 @@ async fn execute<'a, 'b>(
     match rpc.as_str() {
         "has" => return to_js(do_has(txn.read().await.as_read(), from_js(data)?).await),
         "get" => return to_js(do_get(txn.read().await.as_read(), from_js(data)?).await),
-        "scan" => return to_js(do_scan(txn.read().await.as_read(), from_js(data)?).await),
+        "scan" => {
+            return to_js(do_scan(txn.read().await.as_read(), from_js(data.clone())?, data).await)
+        }
         _ => (),
     }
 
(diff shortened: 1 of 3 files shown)


src/embed/connection.rs
```diff
@@ -461,13 +464,24 @@ async fn do_get(read: db::Read<'_>, req: GetRequest) -> Result<GetResponse, Stri
...
+    let receiver: Function = Reflect::get(&req_raw, &JsValue::from_str("receiver"))
+        .map_err(|_| MissingReceiver)?
+        .dyn_into()
+        .map_err(|_| InvalidReceiver)?;
+
     for pe in read.scan((&req.opts).into()) {
-        res.push(ScanItem::try_from(pe).map_err(to_debug)?);
+        let key = JsValue::from_str(std::str::from_utf8(pe.key).unwrap());
+        let val = unsafe { Uint8Array::view(pe.val) };
+        // TODO: receiver can return to us whether to keep going!
```
phritz, line 481: Can you elaborate on this comment? Like what if the receiver decides to do a scan?

src/embed/connection.rs
```diff
@@ -461,13 +464,24 @@ async fn do_get(read: db::Read<'_>, req: GetRequest) -> Result<GetResponse, Stri
...
+        receiver.call2(&JsValue::null(), &key, &val).unwrap();
```
phritz, line 482: Ignoramus here, is it always the case that the receiver is a free function or has some other way of not needed or getting its this/self? Is there any way to misuse it?

tests/wasm.rs
```diff
@@ -359,6 +360,8 @@ async fn test_get_put_del() {
     assert_eq!(dispatch::<_, String>(db, "close", "").await.unwrap(), "");
 }
 
+/*
+TODO: Figure out how to re-enable this test. We still have coverage at JS SDK level luckily.
```
phritz, line 364: Can you add a sentence to say what the issue is?

tests/wasm.rs
```diff
@@ -359,6 +360,8 @@ async fn test_get_put_del() {
     assert_eq!(dispatch::<_, String>(db, "close", "").await.unwrap(), "");
 }
 
+/*
```
phritz, line 363: Probably better to `#[ignore]` it that way we the code doesn't rot and we get a reminder that it is ignored.

phritz (commented):
LGTM assuming there is a way to prevent scan receiver from doing something that deadlocks the system. (Documentation might be an ok solution.)

src/embed/connection.rs
aboodman: I was saying that the receiver could return a boolean to tell us whether to get the next value, and we could potentially simplify batching/limits.

src/embed/connection.rs
aboodman: It's basically part of the interface that the receiver is a free function. Another of way of saying this is that the interface we provide is you send us a function we call with a null `this`. If you want some other `this` you must arrange for that. This is easy and standard in JS (via `bind()`).

PR #208, opened 2020-10-09
Implement Garbage Collection

### LINKED ISSUE #34: Implement refcounting / gc
We now have enough of the components in place to implement refcounting and gc in the dag layer. See https://github.com/rocicorp/repc/issues/3#issuecomment-648711429 for proposed design.

arv:
We use ref counting to implement garbage collection. This is done at the
`dag::Write` layer. We do the garbage collection in `dag::Write::commit`.

- A head keeps a chunk alive.
- A ref from one alive chunk to another chunk keeps the other chunk
  alive.

When we call `set_head` we record the new and old hash so that we can
increment and decrement the ref counts later in `commit`.

When we call `put_chunk` we record the hash so that we can check for
orphaned chunks in `commit`.

Fixes #34

### DIFF
--- a/src/dag/write.rs
+++ b/src/dag/write.rs
@@ -6,8 +6,8 @@ use async_recursion::async_recursion;
 use async_std::sync::RwLock;
 use futures::future::try_join_all;
 use futures::try_join;
-use std::collections::{HashMap, HashSet};
 use std::collections::hash_map::Entry;
+use std::collections::{HashMap, HashSet};
 use std::convert::TryInto;
 use str_macro::str;
 
@@ -79,10 +79,13 @@ impl<'a> Write<'_> {
             Entry::Occupied(mut entry) => {
                 // Keep old if occupied.
                 entry.get_mut().new = hash.map(str::to_string);
-            },
+            }
             Entry::Vacant(entry) => {
-                entry.insert(HeadChange{new: hash.map(str::to_string), old: old_hash});
-            },
+                entry.insert(HeadChange {
+                    new: hash.map(str::to_string),
+                    old: old_hash,
+                });
+            }
         }
 
         Ok(())
--- a/src/dag/write.rs
+++ b/src/dag/write.rs
@@ -1,15 +1,35 @@
-use super::chunk::Chunk;
 use super::key::Key;
+use super::{chunk::Chunk, meta_generated::meta};
 use super::{read, Result};
 use crate::kv;
+use async_recursion::async_recursion;
+use async_std::sync::RwLock;
+use futures::future::try_join_all;
+use futures::try_join;
+use std::collections::HashSet;
+
+#[derive(Debug, Default)]
+struct HeadChange(Option<String>, Option<String>);
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum RefCount {
+    Inc = 1,
+    Dec = -1,
+}
 
 pub struct Write<'a> {
     kvw: Box<dyn kv::Write + 'a>,
+    changed_heads: RwLock<Vec<HeadChange>>,
+    mutated_chunks: RwLock<HashSet<String>>,
 }
 
 impl<'a> Write<'_> {
     pub fn new(kvw: Box<dyn kv::Write + 'a>) -> Write {
-        Write { kvw }
+        Write {
+            kvw,
+            changed_heads: Default::default(),
+            mutated_chunks: Default::default(),
+        }
     }
 
     pub fn read(&self) -> read::Read {
@@ -17,28 +37,47 @@ impl<'a> Write<'_> {
     }
 
     pub async fn put_chunk(&mut self, c: &Chunk) -> Result<()> {
+        // TODO: These can be done in parallel.
+
         self.kvw
             .put(&Key::ChunkData(c.hash()).to_string(), c.data())
             .await?;
+
         if let Some(meta) = c.meta() {
             self.kvw
                 .put(&Key::ChunkMeta(c.hash()).to_string(), meta)
                 .await?;
         }
+
+        self.mutated_chunks
+            .write()
+            .await
+            .insert(c.hash().to_string());
+
         Ok(())
     }
 
     pub async fn set_head(&mut self, name: &str, hash: Option<&str>) -> Result<()> {
+        let old_hash = self.read().get_head(name).await?;
+
+        // TODO: These can be done in parallel.
+
+        let head_key = Key::Head(name).to_string();
         match hash {
-            None => Ok(self.kvw.del(&Key::Head(name).to_string()).await?),
-            Some(h) => Ok(self
-                .kvw
-                .put(&Key::Head(name).to_string(), h.as_bytes())
-                .await?),
+            None => self.kvw.del(&head_key).await?,
+            Some(h) => self.kvw.put(&head_key, h.as_bytes()).await?,
         }
+
+        self.changed_heads
+            .write()
+            .await
+            .push(HeadChange(hash.map(str::to_string), old_hash));
+
+        Ok(())
     }
 
     pub async fn commit(self) -> Result<()> {
+        self.collect_garbage().await?;
         Ok(self.kvw.commit().await?)
     }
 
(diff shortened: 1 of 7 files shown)


aboodman:
https://i.imgflip.com/1qx7x7.jpg

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
 use super::{read, Result};
 use crate::kv;
+use async_recursion::async_recursion;
+use async_std::sync::RwLock;
+use futures::future::try_join_all;
+use futures::try_join;
+use std::collections::HashSet;
+
+#[derive(Debug, Default)]
+struct HeadChange(Option<String>, Option<String>);
```
aboodman, line 12: [taste] I pretty much always regret using tuples for anything. Normal struct?

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum RefCount {
```
aboodman, line 15: [taste] Maybe `RefChange`?

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+    Inc = 1,
```
aboodman, line 16: [taste] Generally, Rust seems to eschew abbreviation. It's the anti-Go.

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+    Dec = -1,
+}
 
 pub struct Write<'a> {
     kvw: Box<dyn kv::Write + 'a>,
+    changed_heads: RwLock<Vec<HeadChange>>,
+    mutated_chunks: RwLock<HashSet<String>>,
```
aboodman, line 23: todo: this is unexpected

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum RefCount {
+    Inc = 1,
+    Dec = -1,
+}
 
 pub struct Write<'a> {
     kvw: Box<dyn kv::Write + 'a>,
+    changed_heads: RwLock<Vec<HeadChange>>,
```
aboodman, line 22: todo: is this changed, or just touched?

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+    mutated_chunks: RwLock<HashSet<String>>,
 }
 
 impl<'a> Write<'_> {
     pub fn new(kvw: Box<dyn kv::Write + 'a>) -> Write {
-        Write { kvw }
+        Write {
+            kvw,
+            changed_heads: Default::default(),
```
aboodman, line 30: [taste]: I think more clear to just construct the write value, unless there is some reason for the genericism of `Default` necessary.

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
-            None => Ok(self.kvw.del(&Key::Head(name).to_string()).await?),
-            Some(h) => Ok(self
-                .kvw
-                .put(&Key::Head(name).to_string(), h.as_bytes())
-                .await?),
+            None => self.kvw.del(&head_key).await?,
+            Some(h) => self.kvw.put(&head_key, h.as_bytes()).await?,
         }
+
+        self.changed_heads
```
aboodman, line 71: What happens if the same head gets set twice in the same tx? Seems like you want a map by head name.

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+            .write()
+            .await
+            .push(HeadChange(hash.map(str::to_string), old_hash));
+
+        Ok(())
     }
 
     pub async fn commit(self) -> Result<()> {
+        self.collect_garbage().await?;
```
aboodman, line 80: I really dislike the error handling in this module, we should fix it soon.

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+        // TODO: Getting the write lock could be done in parallel too.
+
+        if update_mutated_chunks {
+            let mut mutated_chunks = self.mutated_chunks.write().await;
+            mutated_chunks.remove(hash);
+        }
+        Ok(())
+    }
+
+    async fn collect_garbage(&self) -> Result<()> {
```
aboodman, line 156: [taste] I think it reads better to organize the functions roughly top-down, and Rust doesn't care one way or other.

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
+        // We increment all the ref counts before we do all the decrements. This
+        // is so that we do not remove an item that goes from 1 -> 0 -> 1
+
+        let changed_heads = self.changed_heads.read().await;
```
aboodman, line 160: This is cool. FWIW, this pattern would still work even if you used normal struct for changed_heads (you can destructure plain old structs too).

src/dag/write.rs
```diff
@@ -1,55 +1,195 @@
...
         Ok(self.kvw.commit().await?)
     }
 
     #[allow(dead_code)]
     pub async fn rollback(self) -> Result<()> {
         Ok(self.kvw.rollback().await?)
     }
+
+    async fn set_ref_count(&self, hash: &str, count: u16) -> Result<()> {
+        // Ref count is represented as a u16 stored as 2 bytes using BE.
```
aboodman, line 90: [question] did you consider using flatbuffers for the single value? Kinda silly but (a) it takes care of the serialization for you, and (b) is extensible (not sure what we'd use that for but you never know).

aboodman (commented):
This looks fantastic.

I'm concerned about performance since this is such an obviously sensitive thing. Can you please run the benchmark and see the situation? Also, would you mind trying an iteration where we put the counts in the meta and see how that fares?

The far other side of the spectrum is to put the counts in a single key. If values average 1k, then a full replicache will have 20k entries, and the count value will be ~300kb. Not so bad. But to flip a single bit it's a lot of data to move.

The middle (and probably correct) ground is probably something like a b-tree.

aboodman:
But to be clear let's just try the two options for now: put this in the meta key, or put it on its own.

src/dag/write.rs
arv: Anything that was `put_chunked`. We do not check for no op puts

src/dag/write.rs
arv: I had it as a struct but `unzip` wants a tuple so I changed it to that to reduce the amount of work needed

src/dag/write.rs
aboodman: Crap, the todo was a todo to myself to figure this out before sending the review. Nevermind this comment.

src/dag/write.rs
aboodman: nevermind this comment.

src/dag/write.rs
arv: less code duplication

src/dag/write.rs
arv: Nope, we need to keep track of that. Inc, Dec, Inc for example. In practice I don't think we will be setting head more than once per commit but this way we are sure that we do not end up clobbering old head changes

src/dag/write.rs
arv: I did consider it but feels like a lot of code for storing a single number

aboodman:
Sorry now that I think about this a few more moments:

- We probably shouldn't expect to see much perf difference either way right now because we only have a few chunks per commit.
- We (well, I) already decided we didn't want to put things that can mutate in the `meta` area. But maybe this is a reason being so strict doesn't make sense.
- We could maybe get a sense of the perf difference by adding a new benchmark that just writes a single (random) key in its own transaction. In that case, if the difference between reading many small keys in idb vs one bigger key is meaningful, it should show up.

src/dag/write.rs
arv: I see that I ended up with the `map` function after all. Will switch back to a struct.

src/dag/write.rs
aboodman: I don't understand. If we do:

```
set_head("foo", 0x1);
set_head("foo", 0x2);
```

Can't we just run the algorithm with `old = <original value of "foo", before this tx>` and `new = 0x2`?

src/dag/write.rs
aboodman: Agree. I believe every platform we run on is LE. Is there any value to sticking with that?

src/dag/write.rs
arv: I used to detect when the set_head was a no op, but it is very unlikely and the system works fine even if that was the case.

arv:
Size numbers:

**master**:
```
replicache_client.js: 24484
replicache_client.js.br: 4532
replicache_client_bg.wasm: 527419
replicache_client_bg.wasm.br: 152207
```

**ref-counting**:
```
replicache_client.js: 24484
replicache_client.js.br: 4548
replicache_client_bg.wasm: 549671
replicache_client_bg.wasm.br: 158369
```

Perf numbers:

**master**:
```
populate 1024x1000 (clean) x 0.80 MB/s ±0.0% (0 runs sampled)
populate 1024x1000 (dirty) x 0.80 MB/s ±0.0% (0 runs sampled)
scan 1024x1000 x 20.57 MB/s ±0.0% (0 runs sampled)
scan 1024x5000 x 21.15 MB/s ±0.0% (0 runs sampled)
```

**ref-counting**:
```
populate 1024x1000 (clean) x 0.75 MB/s ±0.0% (0 runs sampled)
populate 1024x1000 (dirty) x 0.74 MB/s ±0.0% (0 runs sampled)
scan 1024x1000 x 19.28 MB/s ±0.0% (0 runs sampled)
scan 1024x5000 x 21.06 MB/s ±0.0% (0 runs sampled)
```

arv:
Not clear what the clippy error is about. Does not repro locally.

src/dag/write.rs
arv: no preference between LE or BE. I just picked one... I'll change to LE for simplicity

src/dag/write.rs
arv: I think you are right... just need to figure out how to get the original value. I could get it the first time `set_head("foo")` is called and just update the following `set_head("foo")` calls to not overwrite the old... maybe there is a way to get the underlying value from the kv?

arv:
I'll put the count in meta next week. Based on my earlier work it is a lot more complicated...

src/dag/write.rs
aboodman: You can get it on first write and store it. I don't think it makes sense to expose it from kv::Write.

aboodman:
You have to run clippy with warnings.

arv:
image

arv:
I created another branch where the ref count is in the meta data.


Perf:
```
populate 1024x1000 (clean) x 0.77 MB/s ±0.0% (0 runs sampled)
populate 1024x1000 (dirty) x 0.71 MB/s ±0.0% (0 runs sampled)
scan 1024x1000 x 19.15 MB/s ±0.0% (0 runs sampled)
scan 1024x5000 x 21.58 MB/s ±0.0% (0 runs sampled)
```

https://github.com/arv/repc/pull/1

phritz (commented):
Wow, this is really clean. Nice.

src/dag/write.rs
```diff
@@ -1,55 +1,218 @@
...
+            self.remove_all_related_keys(hash, true).await?;
+        } else {
+            self.set_ref_count(hash, new_count as u16).await?;
+        }
+
+        Ok(())
+    }
+
+    async fn set_ref_count(&self, hash: &str, count: u16) -> Result<()> {
+        // Ref count is represented as a u16 stored as 2 bytes using BE.
```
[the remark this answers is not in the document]
arv: haha... was BE at one point... fixing

src/dag/write.rs
```diff
@@ -1,55 +1,218 @@
...
+                        self.change_ref_count(r, rc).await?;
+                    }
+                }
+            }
+        }
+
+        if new_count == 0 {
+            self.remove_all_related_keys(hash, true).await?;
+        } else {
+            self.set_ref_count(hash, new_count as u16).await?;
```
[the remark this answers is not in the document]
arv: I wasn't sure either. Seems like 65k incoming refs is pretty high. I was considering making this "adjustible", meaning that we use `u8`, `u16` and `u32` as needed.

Also, u32 would be ok. It is just 2 more bytes than u16

PR #209, opened 2020-10-13
Make original_hash weak

arv:
The target is kept alive by the different named heads.

https://github.com/rocicorp/repc/issues/105


aboodman:
lgtm

PR #211, opened 2020-10-13
Parallelize put_chunk

arv:
There is no need to write data, meta and mutated_chunks serially

### DIFF
--- a/src/dag/write.rs
+++ b/src/dag/write.rs
@@ -5,6 +5,7 @@ use crate::kv;
 use async_recursion::async_recursion;
 use async_std::sync::RwLock;
 use futures::future::try_join_all;
+use futures::future::TryFutureExt;
 use futures::try_join;
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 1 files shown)


src/dag/write.rs
```diff
@@ -43,22 +44,25 @@ impl<'a> Write<'_> {
...
-                .put(&Key::ChunkMeta(c.hash()).to_string(), meta)
-                .await?;
-        }
-
-        self.mutated_chunks
-            .write()
-            .await
-            .insert(c.hash().to_string());
+        let data_key = Key::ChunkData(c.hash()).to_string();
+        try_join!(
```
aboodman, line 48: I'm shocked you got this to work. Nice.

aboodman (commented):
LGTM

src/dag/write.rs
arv: The thing that made it easier was to extract the conditional one into a function at first... then I realized async blocks allows a concise syntax

aboodman:
lgtm

PR #212, opened 2020-10-13
First cut at indexes

aboodman:


src/db/index.rs
```diff
@@ -0,0 +1,452 @@
...
+        test(
+            "k",
+            &[] as &[u8],
+            "/",
+            Err(DeserializeError(
+                "EOF while parsing a value at line 1 column 0".to_string(),
+            )),
+        );
+        // no matching target
+        test("k", "{}".as_bytes(), "/foo", Ok(vec![]));
```
arv, line 303: FYI `b"{}"`

src/db/read.rs
```diff
@@ -79,11 +105,52 @@ impl<'a> Read<'a> {
...
+            }
+        }
+        match opts.index_name {
+            Some(name) => {
+                let idx = self
+                    .indexes
+                    .get(name)
+                    .ok_or_else(|| UnknownIndexName(name.to_string()))?;
+                let guard = idx.get_map(&self.dag_read).await.map_err(GetMapError)?;
+                send(guard.get_map(), opts, callback);
```
arv, line 140: Kind of strange that we do `get_map` and then `get_map` again?

Maybe on of them should be renamed?

src/db/write.rs
```diff
@@ -120,21 +125,62 @@ impl<'a> Write<'a> {
...
+            .map_err(AddNewIndexEntriesError)?;
+        self.map.put(key, val);
+        Ok(())
     }
 
-    pub fn del(&mut self, key: Vec<u8>) {
-        let old_val = self.map.get(&key);
+    pub async fn del(&mut self, key: Vec<u8>) -> Result<(), DelError> {
+        use DelError::*;
+        // boooo copy (needed to appease borrowck)
```
arv, line 148: https://github.com/rocicorp/repc/pull/219

arv (commented):
Not quite done

src/db/write.rs
```diff
@@ -159,8 +269,17 @@ impl<'a> Write<'a> {
             .flush(&mut self.dag_write)
             .await
             .map_err(FlushError)?;
-
-        let commit = match self.meta {
+        let mut index_metas = Vec::new();
+        for (_, index) in self.indexes.into_iter() {
```
arv, line 273: use `.values()`?

src/db/write.rs
```diff
@@ -249,7 +403,9 @@ mod tests {
         )
         .await
         .unwrap();
-        w.put("foo".as_bytes().to_vec(), "bar".as_bytes().to_vec());
+        w.put("foo".as_bytes().to_vec(), "bar".as_bytes().to_vec())
```
arv, line 406: b"foo"

src/embed/connection.rs
```diff
@@ -470,32 +472,57 @@ async fn do_scan(
...
+        .await
+        .map_err(DBError)?;
+    Ok(CreateIndexResponse {})
+}
+
+async fn do_drop_index(
+    write: &mut db::Write<'_>,
+    req: DropIndexRequest,
+) -> Result<DropIndexResponse, DropIndexError> {
+    use DropIndexError::*;
```
arv, line 521: `use` does not seem worth it for a single use

src/embed/types.rs
```diff
@@ -132,6 +132,8 @@ pub struct ScanOptions {
     pub prefix: Option<String>,
     pub start: Option<ScanBound>,
     pub limit: Option<u64>,
+    #[serde(rename = "indexName")]
```
arv, line 135: I know it is too late but i wish we had used `snake_case` for our JSON

arv (commented):
LGTM

src/db/read.rs
aboodman: Heh yeah, didn't notice that. Before it was `to_map` but Clippy bitched at me because there's a convention that `to_*` have a particular signature.

src/db/write.rs
```diff
@@ -146,6 +192,70 @@ impl<'a> Write<'a> {
...
+    pub async fn create_index(
+        &mut self,
+        name: &str,
+        key_prefix: &[u8],
+        json_pointer: &str,
+    ) -> Result<(), CreateIndexError> {
+        use CreateIndexError::*;
+
+        let definition = commit::IndexDefinition {
+            name: name.to_string(),
```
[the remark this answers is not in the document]
aboodman: Yeah I think so - thats my normal style, I'm not sure why I didn't do it here. I was really tired and moving fast.

src/db/write.rs
```diff
@@ -146,6 +192,70 @@ impl<'a> Write<'a> {
...
+            },
+        ) {
+            index::index_value(
+                &mut index_map,
+                index::IndexOperation::Add,
+                entry.key,
+                entry.val,
+                json_pointer,
+            )
+            .map_err(|e| IndexError((name.to_string(), entry.key.to_vec(), e)))?;
```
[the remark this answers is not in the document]
aboodman: Yep.

src/embed/connection.rs
aboodman: Kind of disagree here. I like there to be conventions in the code so that you don't have to think about each special case. Also makes it easier to change code if it all follows the same pattern.

(I actually have an idea for a macro to automate some of this if I ever have spare time)

src/embed/types.rs
aboodman: It's not too late! This is private API. We can just change it and change the client in lockstep.

aboodman: (and I agree)

phritz (commented):
Really nice how indexing just fits right in to the design -- scan works the same for both index and normal queries, gcing old indexes just works etc.

Main observation is that the number of TODOs in the code makes me worried that since we are not done done with indexing that we risk forgetting to do some important things by the time we think we are ready for "launch." Suggest we move the TODOs into an issue that says 'finish up indexing', or pointers to the TODOs, and include it in the current milestone.

A couple of other maybe-outstanding things to get this done done:
- did we get a picture of the scan performance improvement indexing gets us? i think i saw something in slack that the impact to write, but i didn't see the speedup for reads. 
- related: do our perf benchmarks have the right set of index-related tests?
- customer-facing docs need to be updated
- integration: i assume you're working with cron helping them add indexing?

src/db/write.rs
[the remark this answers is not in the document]
aboodman: In this situation create_index actually ends up needing to make several copies so I don't think it's as clear. But yeah generally I think that if a function needs to take an owned copy of something, it should make that clear by requesting that the callee pass it one.

src/db/index.rs
aboodman: https://github.com/rocicorp/repc/pull/239

src/db/read.rs
aboodman: I'm not really sure what would be better here so leaving it.

src/db/write.rs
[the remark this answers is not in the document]
aboodman: https://github.com/rocicorp/repc/pull/239

src/db/write.rs
aboodman: This is weird but `self.indexes.into_iter()` iterates `(String, Index)` (by value) but `self.indexes.values().into_iter()` iterates `(&str, &Index)` (by reference).

src/db/write.rs
aboodman: done

src/embed/types.rs
aboodman: This makes more sense as a separate PR. Agree we should do it but not doing it as part of this.

Issue #124, opened 2020-08-30
disallow putting empty key into the map


phritz:
the key has to be non-empty, else we can't distinguish removing it from a top-level remove in patch. a top level remove is `{"op": "remove", "path": "/"}`, and so would be removing the empty key.

arv:
It is not clear to me if we need to disable the key `""`. I looked at the specs for JSON Patch and JSON Path and it seems like the path `""` is the whole document and the path `"/"` is the value of the key `""`. I will see what happens in our code.

arv:
I did some more research. `""` is clearly the root. Normally it makes no sense to remove the root, replacing the root works in some of the online demos I found though.

Are we relying on `"/"` to mean clear the db? I see this https://github.com/rocicorp/repc/blob/master/src/sync/patch.rs#L36 but are we generating this? Can we use `{"op": "replace", "path": "", "value": {}}` instead?

phritz:
The diff of the noms maps doesn't ever generate a top-level remove of "/" but diffserver [returns it](https://github.com/rocicorp/diff-server/blob/b7795834dabecc13e6c54ce761efa9f9612202c6/db/diff.go#L15) in [several cases](https://github.com/rocicorp/diff-server/blob/b7795834dabecc13e6c54ce761efa9f9612202c6/db/diff.go#L48), eg if it doesn't have a copy of their base state. 

To answer your question yes we are relying on remove "/" to clear the db. This is different than what json path would suggest, which as you say is remove "". Our diff and patch came from noms so my guess is there was some reason lost to time that we [required patch keys to start with "/"](https://github.com/rocicorp/diff-server/blob/bf27412b085fb8fba1b3d4393e0f7d199fbb2cbd/util/noms/jsonpatch/jsonpatch.go#L121) and it just got carried over into the diffserver. 

Seems like the right resolution here is to fix this patch syntax we use instead of disallowing empty key.

arv:
I see that we have a `version` param. Maybe this is the right time to use version `2`?

https://github.com/rocicorp/repc/blob/740b0482c69ee3160d0dce3a9cb5770db06eb084/src/sync/mod.rs#L119

phritz:
Yes. Though we haven't really settled on a general strategy, this is more of a special case. https://github.com/rocicorp/repc/issues/119 covers a few related things and I'm supposed to make a proposal at some point.

arv:
Here is the dance:

[...]

Issue #119, opened 2020-08-26
we need to version our rpcs


phritz:
fixing https://github.com/rocicorp/diff-server/issues/54 without versioning (eg https://github.com/rocicorp/diff-server/pull/55) is a breaking change to existing clients. we should solve this issue the same way for all our rpcs, including between cli and the account service. note there we have slightly more structure, namely a wrapper: https://github.com/rocicorp/account-service/blob/e07de54c1cf601ecb01b743ddd8078061333821d/rpc/wrapper.go#L34. 

aaron and i have talked in the past about solving this problem eg with protos and agreed to wait until the need is more pressing. filing this now as that time is approaching rapidly.

notes:
- for right now to fix the proximate issue the proposal is to introduce a one-off version field into the pullrequest. 
- the diff-server gets pushed from head of master, but even if it didn't we would still need versioning else our releases would be gated by our slowest-to-update client.

aboodman:
I agree that adding a version field to the pull request nicely solves this problem. Are you advocating for doing more work right now than just fixing the proximate issue? If so what are you advocating for?

phritz:
> Are you advocating for doing more work right now than just fixing the proximate issue?

No. I'm saying "let's solve this problem soon" because we are to the point where we want to be careful about what we break and when. I'm not proposing any particular solution.

aboodman:
On Wed, Aug 26, 2020 at 8:20 AM Phritz <notifications@github.com> wrote:

> Are you advocating for doing more work right now than just fixing the
> proximate issue?
>
> No. I'm saying "let's solve this problem soon" because we are to the point
> where we want to be careful about what we break and when. I'm not proposing
> any particular solution.
>
OK - I don't have any ideas for anything more general or feel very
compelled to find them (because I don't feel worried about being able to do
it when the time comes, or hopeful about finding a fully generalized
solution). But if you do, go for it.

aboodman:
On Wed, Aug 26, 2020 at 8:25 AM Aaron Boodman <aaron@aaronboodman.com>
wrote:

>
>
> On Wed, Aug 26, 2020 at 8:20 AM Phritz <notifications@github.com> wrote:
>
>> Are you advocating for doing more work right now than just fixing the
>> proximate issue?
>>
>> No. I'm saying "let's solve this problem soon" because we are to the
>> point where we want to be careful about what we break and when. I'm not
>> proposing any particular solution.
>>
> OK - I don't have any ideas for anything more general or feel very
[...]

PR #213, opened 2020-10-13
JSON patch compatible clear

arv:
This adds special code for dealing with
`{"op":"replace","path":"","value":{}}` which is what we use to clear
the map.

We still treat `{"op":"remove","path":"/"}` as clear for now but we
should update the other layers to not use that.

Towards #124

### DIFF
--- a/src/sync/patch.rs
+++ b/src/sync/patch.rs
@@ -18,6 +18,16 @@ pub struct Operation {
 pub fn apply(db_write: &mut db::Write, patch: &[Operation]) -> Result<(), PatchError> {
     use PatchError::*;
     for op in patch.iter() {
+        // Special case `{"op": "replace", "path": "", "value": {}}` which means
+        // replace the root with a new map, in other words, clear the map.
+        if op.path.is_empty() {
+            if op.op == OP_REPLACE && op.value_string == "{}" {
+                db_write.clear();
+                continue;
+            }
+            return Err(InvalidPath(op.path.clone()));
+        }
+
         let mut chars = op.path.chars();
         if chars.next() != Some('/') {
             return Err(InvalidPath(op.path.clone()));
@@ -31,12 +41,7 @@ pub fn apply(db_write: &mut db::Write, patch: &[Operation]) -> Result<(), PatchE
             }
             // Should we error if we try to remove a key that doesn't exist?
             OP_REMOVE => {
-                if key.is_empty() {
-                    // Top-level remove (path == "/").
-                    db_write.clear();
-                } else {
-                    db_write.del(key);
-                }
+                db_write.del(key);
             }
             _ => return Err(InvalidOp(op.op.to_string())),
         };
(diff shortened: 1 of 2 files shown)


aboodman (approved):

aboodman:
Thanks for fixing my crappy code.

src/sync/patch.rs
```diff
@@ -18,6 +18,16 @@ pub struct Operation {
...
     use PatchError::*;
     for op in patch.iter() {
+        // Special case `{"op": "replace", "path": "", "value": {}}` which means
+        // replace the root with a new map, in other words, clear the map.
+        if op.path.is_empty() {
+            if op.op == OP_REPLACE && op.value_string == "{}" {
+                db_write.clear();
+                continue;
+            }
+            return Err(InvalidPath(op.path.clone()));
```
phritz, line 28: Not really an invalid path, more like an invalid value.

phritz (approved):
Let's talk about how to roll this out. 

We can't change the diffserver to stop sending the old remove and start sending the new replace because it deploys automatically from master, and that would break existing clients in the wild. If we don't want to break old clients but also want to support new clients we need a signal from the client about which way to remove. As suggested on issue, the mechanism we have atm is the version field. I think for now we should:
1. (repc) Change client to send version == 3.
2. (diffserver) Change diffserver to return the old remove if version is 2, or the new remove if version is 3. Merge that change which releases it.
3. (repc) Remove support for the old remove from repc.
4. (repc) As soon as we do the next release of repc, tell our customers to upgrade to it.
5. (diffserver) After they have or some period of time remove support for the old remove from diffserver and error out if client sends version < 3, and release that.

An alternative, simpler flow that breaks things temporarily is is:
1. (repc) Remove old remove support from repc and release repc. (This breaks repc until step 2 completes, hopefully quickly.)
2. (diffserver) Remove old support from diffserver and have it return the new replace, merge the change which releases it. (This breaks old clients, they now need to upgrade.)
3. Tell customers their clients are broken and to upgrade to the new repc.

(Not having to have this conversation every time we have a breaking change is one of the main points of https://github.com/rocicorp/repc/issues/119, btw).

arv:
I'm in favor for the more elaborate dance.

phritz:
> I'm in favor for the more elaborate dance.

yeah at some point soon we *must* be doing that dance at all times, might as well start now

src/sync/patch.rs
```diff
@@ -18,6 +18,16 @@ pub struct Operation {
 pub fn apply(db_write: &mut db::Write, patch: &[Operation]) -> Result<(), PatchError> {
     use PatchError::*;
     for op in patch.iter() {
+        // Special case `{"op": "replace", "path": "", "value": {}}` which means
```
phritz, line 21: > `{"op": "replace", "path": "", "value": {}}`

Should be:

`{"op": "replace", "path": "", "valueString": "{}"}`

arv:
diff-server got pushed... committing

PR #214, opened 2020-10-14
Flatten refs iter

arv:
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten

"This is useful when you have an iterator of iterators or an iterator of
things that can be turned into iterators and you want to remove one
level of indirection."

### DIFF
--- a/src/dag/chunk.rs
+++ b/src/dag/chunk.rs
@@ -38,16 +38,18 @@ impl Chunk {
         &(self.data.0[self.data.1..])
     }
 
-    // TODO: It would be nice to flatten the option down into an empty
-    // iterator for caller convenience, but could not find a zero-alloc
-    // way to do that.
-    pub fn refs(&self) -> Option<impl Iterator<Item = &str>> {
-        if let Some(buf) = self.meta() {
+    pub fn refs(&self) -> impl Iterator<Item = &str> {
+        let inner = if let Some(buf) = self.meta() {
             if let Some(refs) = meta::get_root_as_meta(buf).refs() {
-                return Some(refs.iter());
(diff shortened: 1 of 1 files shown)


src/dag/chunk.rs
```diff
@@ -38,16 +38,18 @@ impl Chunk {
         &(self.data.0[self.data.1..])
     }
 
-    // TODO: It would be nice to flatten the option down into an empty
-    // iterator for caller convenience, but could not find a zero-alloc
-    // way to do that.
-    pub fn refs(&self) -> Option<impl Iterator<Item = &str>> {
-        if let Some(buf) = self.meta() {
+    pub fn refs(&self) -> impl Iterator<Item = &str> {
+        let inner = if let Some(buf) = self.meta() {
```
aboodman, line 42: I think an even smaller way to do this is with `flat_map`. Something like:

```
self.meta().iter().flat_map(|buf| meta::get_root_as_meta(buf).refs()).iter().flat_map(|refs| refs.iter())
```
?

src/dag/chunk.rs
```diff
@@ -73,13 +75,9 @@ impl Chunk {
...
 impl PartialEq for Chunk {
     fn eq(&self, other: &Self) -> bool {
-        match self.refs() {
-            None => other.refs().is_none(),
-            Some(s) => match other.refs() {
-                None => false,
-                Some(o) => s.eq(o),
-            },
-        }
+        // TODO(arv): Compare data too?
```
aboodman, line 78: eep, definitely. Or at least hash.

aboodman (approved):

PR #216, opened 2020-10-15
Use binary search for has and get

arv:
First look in the pending, then binary search the flatbuffers::Vector.

The std binary_search does not work on flatbuffers::Vector (at least I
could not figure out how to do it without a copy) so I copied and
modified the code.

### DIFF
--- a/src/prolly/leaf.rs
+++ b/src/prolly/leaf.rs
@@ -1,7 +1,8 @@
-use super::leaf_generated::leaf;
+use super::leaf_generated::leaf::{self, LeafEntry};
 use super::Entry;
 use crate::dag::Chunk;
 use flatbuffers::FlatBufferBuilder;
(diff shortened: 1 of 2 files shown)


arv:
Got side tracked... but I had some fun!

aboodman:
Nice but I kind of go back and forth on this. Once we have prolly trees then we should only have a small number of items in each node. So due to cpu caching I would be surprised if this makes it any faster or could even make it worse. It also increases complexity significantly

I kinda feel like in Rust we want to conserve our complexity budget wherever we can 😬. @phritz ?

phritz:
Generally in favor of less complexity in rust. In this case seems like it should improve performance for now and the complexity is at least concentrated and doesn't have to be understood to use the thingy, so in favor of merging this.

aboodman:
OK, LGTM then

PR #217, opened 2020-10-15
Phritz index/1 - DO NOT MERGE

phritz:
address two indexing TODOs:
- implement dropIndex
- make createIndex w/same definition a nop

i went further than asked in the second one and made replacing an existing index with an index of a different definition an error. easy enough to back out if undesirable.


src/db/write.rs
```diff
@@ -193,8 +193,18 @@ impl<'a> Write<'a> {
         key_prefix: &[u8],
         json_pointer: &str,
     ) -> Result<(), CreateIndexError> {
-        // TODO: This should be a nop if the specified index already exists.
         use CreateIndexError::*;
+
+        // Check to see if the index already exists.
+        if let Some(index) = self.indexes.get(name) {
+            if index.meta.definition.key_prefix == key_prefix
```
aboodman, line 200: Use `PartialEq` instead so that this generalizes as definition becomes more sophisticated.

src/db/write.rs
```diff
@@ -193,8 +193,18 @@ impl<'a> Write<'a> {
...
+                && index.meta.definition.json_pointer == json_pointer
+            {
+                return Ok(());
+            }
+            return Err(IndexExistsWithDifferentDefinition);
```
aboodman, line 205: Yeah I'm not sure what correct behavior is but this seems like a conservative starting point. Good idea.

aboodman (approved):
Thank you!

src/db/write.rs
phritz: I'm not sure what that would mean in this context. (And I'm also not sure why an index definition would not be reflexive (Eq)).

aboodman: You're right, I was reading this too fast.

aboodman: Take that back, I was right. This is what I meant:

https://github.com/rocicorp/repc/pull/212/commits/773b2319be24d28124850441e85b79831604fdfc

aboodman: (This way if we add fields to definition in the future we won't have to remember to add them to this path).

aboodman:
Thanks again, merged this into the main PR.

PR #221, opened 2020-10-17
Quick fix for sync 🔪 indexes

aboodman:


aboodman:
@phritz (on monday) let's talk about whether this makes sense or we should do something smarter.

src/prolly/leaf.rs
```diff
@@ -83,6 +83,11 @@ impl Leaf {
         }
     }
 
+    pub fn len(&self) -> usize {
+        let root = leaf::get_root_as_leaf(self.chunk.data());
+        root.entries().map(|v| v.len()).unwrap_or(0)
```
arv, line 88: I think it is safe to unwrap `entries()`. We never leave it put.

aboodman: You are right, fixed in https://github.com/rocicorp/repc/pull/239.

PR #223, opened 2020-10-20
First cut at license stuffs

aboodman:
This change licenses repc as Business Source License (BSL).

BSL is a "parameterized" license. It allows usage subject to a set of specified restrictions initially, then falls back to pure open source as of a "change date".

You can read some background on Business Source License here:

https://mariadb.com/bsl-faq-adopting/
https://blog.adamretter.org.uk/business-source-license-adoption/

There are various interesting questions around how to use BSL:

# What should the "Additional Use Grant" be?

I looked for inspiration at what the following projects did and adapted them:

* https://github.com/zerotier/ZeroTierOne/blob/master/LICENSE.txt
* https://mariadb.com/projects-using-bsl-11/

I think we will eventually want to get a lawyer to look at this bit but I think this is OK as a start (we can always tighten it later).

# How far in the future should the "Change Date" be?

Most BSL projects to date use Change Dates that are 3 or 4 years from time of publish. This feels too long to me so I'm proposing 2 years to start.

# How often should the "Change Date" be updated?

Most projects bump the change date with either major or minor versions. I'd ideally like to do it with every commit, but I could not find an example of that in the community so I propose we do [what CockroachDB does](https://www.cockroachlabs.com/blog/oss-relicensing-cockroachdb/) and bump it with minor versions (but not patch versions).


LICENSE
```diff
@@ -0,0 +1,4 @@
+Source code in this repository is variously licensed under the Business Source
```
phritz, line 1: Need same in all our repos I guess.

licenses/BSL.txt
```diff
@@ -0,0 +1,98 @@
+Business Source License 1.1
+
+Parameters
+
+Licensor:             Rocicorp, LLC
+Licensed Work:        repc 0.10.1
+                      The Licensed Work is (c) 2020 Rocicorp, LLC.
+Additional Use Grant: You may use the Licensed Work in a commercial offering
+                      with less than 100 client instances, or in a
+                      non-commercial offering with less than 10,000 client
```
phritz, line 10: 10k clients is very roughly on the order of $100/month to support at the diffserver, not counting any time we spend on them.

licenses/BSL.txt
```diff
@@ -0,0 +1,98 @@
...
+                      instances.
+
+Change Date:          2020-10-20
+
+Change License:       Apache License, Version 2.0
+
+For information about alternative licensing arrangements for the Software,
+please visit: https://cockroachlabs.com/
```
phritz, line 18: Is this the right link?

phritz (commented):
Really like the BSL!

licenses/BSL.txt
aboodman: heh. bound to happen.

LICENSE
aboodman: Yes, just starting here.

aboodman:
This is ready for review now. Check out the tool bump.go

licenses/BSL.txt
aboodman: There is actually a technical distinction here. This license just covers legal access to the *code* -- not to the diff server. That is a different question.

Part of me just feels like the code should be completely free for non-commercial use, but 10k feels like a conservative start.

phritz: > Part of me just feels like the code should be completely free for non-commercial use

That sounds right. We can just charge for non-commercial use of the diffserver that matters, cost-wise. "First 10k free" seems like a fine start.

PR #232, opened 2020-10-21
Stop depending on window()

### LINKED ISSUE #228: Do not use Window
We are using `web_sys::window()` in a bunch of places. This does not work in a worker because there is no `window`/`Window` in a worker.

We should probably just use `js_sys::global()` in combination with `js_sys::reflect::get`

Blocking https://github.com/rocicorp/replicache-sdk-js/issues/143

arv:
Workers do not have a window.

The new preferred way is to use `global_property("indexedDB")` instead
which gets the property of the global object, no matter if that is
`window`, `self`, `globalThis` etc.

Fixes #228

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,9 +74,6 @@ features = [
     "RequestInit",
     "RequestMode",
     "Response",
-    
-    # Do not add Window since we want repc to work in a Worker.
-    # "Window",
 ]
 
 [lib]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,9 @@ features = [
     "RequestInit",
     "RequestMode",
     "Response",
-    "Window",
+    
+    # Do not add Window since we want repc to work in a Worker.
+    # "Window",
 ]
 
 [lib]
--- a/crates/bench/Cargo.toml
+++ b/crates/bench/Cargo.toml
@@ -10,6 +10,8 @@ log = "0.4"
 thousands = "0.2.0"
 wasm-bench-macro = { path = "../bench-macro" }
 web-sys = "0.3.42"
+js-sys = "0.3.40"
+wasm-bindgen = "0.2"
 
 [lib]
 test = false
--- a/crates/bench/Cargo.toml
+++ b/crates/bench/Cargo.toml
@@ -6,12 +6,12 @@ edition = "2018"
 
 [dependencies]
 async-fn = { path = "../async-fn" }
+js-sys = "0.3.40"
 log = "0.4"
 thousands = "0.2.0"
 wasm-bench-macro = { path = "../bench-macro" }
-web-sys = "0.3.42"
-js-sys = "0.3.40"
 wasm-bindgen = "0.2"
+web-sys = "0.3.42"
 
 [lib]
 test = false
--- a/crates/bench/src/lib.rs
+++ b/crates/bench/src/lib.rs
@@ -1,15 +1,24 @@
 use async_fn::AsyncFn1;
+use js_sys::Reflect;
 use log::error;
 use std::cmp;
 use thousands::Separable;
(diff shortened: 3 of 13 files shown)


Cargo.toml
```diff
@@ -74,7 +74,9 @@ features = [
     "RequestInit",
     "RequestMode",
     "Response",
-    "Window",
+    
+    # Do not add Window since we want repc to work in a Worker.
```
aboodman, line 78: I don't think we'll need the comment here since presumably we'll have tests exercising it in sw which will fail...

crates/bench/Cargo.toml
```diff
@@ -10,6 +10,8 @@ log = "0.4"
 thousands = "0.2.0"
 wasm-bench-macro = { path = "../bench-macro" }
 web-sys = "0.3.42"
+js-sys = "0.3.40"
```
aboodman, line 13: sort

crates/bench/src/lib.rs
```diff
@@ -1,15 +1,24 @@
 use async_fn::AsyncFn1;
+use js_sys::Reflect;
 use log::error;
 use std::cmp;
 use thousands::Separable;
 pub use wasm_bench_macro::wasm_bench;
+use wasm_bindgen::JsValue;
+use web_sys::Performance;
+
+pub fn performance() -> Performance {
```
aboodman, line 10: Unrelated but I think maybe we should get rid of this crate ... it's not really the way we've decided to do benchmarking and I'm struggling to understand when we'd use it.

src/fetch/errors.rs
```diff
@@ -8,10 +10,15 @@ pub enum FetchError {
...
-    InvalidResponseFromJS,
-    NoWindow,
+    InvalidResponseFromJs(JsValue),
     RequestFailed(String),
     RequestTimeout(async_std::future::TimeoutError),
     UnableToCreateRequest(String),
     UnableToSetRequestHeader(String),
 }
+
+impl From<JsValue> for FetchError {
```
aboodman, line 20: Do you still need this? Seems like `map_err(InvalidResponseFromJs)` is doing the same thing?

src/util/uuid/mod.rs
```diff
@@ -10,12 +12,10 @@ pub fn uuid() -> String {
 
 #[cfg(target_arch = "wasm32")]
 fn make_random_numbers(numbers: &mut [u8]) {
-    web_sys::window()
-        .expect("window is not available")
-        .crypto()
-        .expect("window.crypto is not available")
+    global_property::<web_sys::Crypto>("crypto")
```
aboodman, line 15: Can you add a TODO/bug to properly handle these errors rather than crashing?

src/fetch/errors.rs
arv: There are some cases of `?` in the code. Maybe it is better to be more explicit?

Also, maybe `InvalidResponseFromJs` was only intended for `Response`?

arv: Ended up removing it and adding more specific errors

aboodman:
lgtm

src/util/uuid/mod.rs
```diff
@@ -10,12 +12,11 @@ pub fn uuid() -> String {
 
 #[cfg(target_arch = "wasm32")]
 fn make_random_numbers(numbers: &mut [u8]) {
-    web_sys::window()
-        .expect("window is not available")
-        .crypto()
-        .expect("window.crypto is not available")
+    // TODO(arv): Return result instead of crashing.
```
phritz, line 15: Please!

phritz (commented):
LGTM

src/util/uuid/mod.rs
arv: https://github.com/rocicorp/repc/issues/248

PR #244, opened 2020-10-29
Remove Option around IdbStore::new

arv:
We were returning a `Result<Option<IdbStore>>` but we never returned
`Ok(None)`... also if we want something like that maybe that should be
an error?


aboodman:
lgtm

PR #245, opened 2020-10-29
Remove local create date

### LINKED ISSUE #109: Implement local_create_date on commits
At first I thought that caller could pass this in and eliminate date code in Rust, but we need this at least for the genesis? So maybe we should find some lightweight rust date code afterall.

arv:
Remove `local_create_date` from commits.

We keep the field in the flatbuffer for backwards compat.

Fixes #109


aboodman (commented):
🎉

phritz (commented):
Yes! thank you!

> We keep the field in the flatbuffer for backwards compat.

think we are still in the time period where you can break it if you want.

arv:
> think we are still in the time period where you can break it if you want.

I agree. I just wanted to hold off on that in this PR.

PR #246, opened 2020-10-29
Experimenting with required flatbuffer fields

arv:
This marks fields in leaf.fbs as required

That has the effect that the API gets a little bit cleaner.


arv:
This should be safe because once marked as required there is no way to leave out the fields.

What about reading data with missing fields? We already asserted that we didn't write empty key/val entries

WDYT?

aboodman:
Let's talk about this in Slack. Fritz and I discussed this previously and
decided it was a bad idea.

Issue #233, opened 2020-10-22
ScanOptions.start and prefix don't work over indexes


aboodman:
Looks like a combination of good news and bad news:

- good news: enum works the way we hoped it did: https://docs.rs/bytekey-fix/0.5.0/bytekey_fix/ser/struct.Serializer.html#enums
- bad news: since scan is not reconstructing this serialization for index keys, scanning from any position other than the start of the index doesn't work

Issue #218, opened 2020-10-16
Indexes followup


aboodman:
- add remainder of unit tests (scan after write, !scan after drop)
- add a few embed tests
- parallelize writes where trivially possible
- delay loading main map (aka "primary index" in db terms) too -- if we protect it by a mutex in the same way we can share code btwn the primary and secondary indexes. also for many use cases, users won't even need the main map, we're loading it for no reason.
- evaluate json pointer more efficiently, w/o decoding entire value

arv:
- [ ] Use iterator for `indexes` - https://github.com/rocicorp/repc/pull/212/files#diff-7ee60ce4b0868c5e59fe278edeed1c89d10c8421fa4372697583952cb2da6591R147

aboodman:
[ ] Do something to prevent nop useless mutations that just add indexes from being sent to the server.

Right now, the only way to add an index is inside a mutation. This mutation doesn't (usually) actually mean anything to the server, so developers have to ignore it. I guess either:

- We need to implement Erik's proposal https://github.com/rocicorp/replicache-sdk-js/issues/144 to have a convenience method for creating indexes, and then implement that by having a special local transaction, and then filter that transaction out of batch requests, or:
- We need to have a special transaction type that only permits local changes?

aboodman:
- [ ] Decide whether https://github.com/rocicorp/repc/pull/221 was the correct solution to persisting indexes across syncs, or whether something else is needed.

aboodman:
I have a draft patch for 'clear' here: https://github.com/rocicorp/repc/pull/229/commits/b7719e6f0ac48f05ff5a69799ec9d5f1ec42b48f. Not going to land it @phritz you can take it as a starting point or not.

phritz:
OK trying to formulate the master list of indexing followups. aaron has a good start above but I'd like to round it out with TODOs from the code and issues filed recently. @aboodman @arv please:
  1. say your opinion on the following open question: [do we reduce scope to only index strings](https://rocicorp.slack.com/archives/G013XFG80JC/p1603300416000600?thread_ts=1603223508.012000&cid=G013XFG80JC)? I say YES. I think we should keep using bytekey in order to facilitate change or expansion of key types.
  1. Look over the list below and tell me what if anything is missing or should be shifted around. Once we have agreement I will replace the top comment with the list and get started. 

**P1: blocks clean, correct indexing**; in rough order of priority
[...]

PR #250, opened 2020-10-31
get index scanning working

phritz:
i'm not super happy with how this turned out. in particular the ScanOptionsInternal business is ugly. i have TODOs in the code to clean it up and will add to 218. if you have ideas/inspiration have at at it.

progress towards #218 
closes https://github.com/rocicorp/repc/issues/233

### DIFF
--- a/src/db/index.rs
+++ b/src/db/index.rs
@@ -88,16 +88,46 @@ pub enum IndexFlushError {
     MapFlushError(prolly::FlushError),
 }
 
-// We only index strings for now.
+// IndexValue describes the type of the secondary index value. If we add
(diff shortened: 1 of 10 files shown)


aboodman:
I was playing with this and believe I may have found a bug: https://github.com/rocicorp/replicache-sdk-js/pull/171/files#r515533580

phritz:
Bah, looks like I didn't test exclusive so would not surprise me if there is a bug there. Will address Monday.

aboodman (commented):
I think it is fine to change the definition of ScanOptions however you want to make it easier to work with. It's not a design constraint that it use references.

PR #253, opened 2020-11-04
simplify ScanOptions

phritz:
major improvement in my mind:
- flatten ScanOptions
- make ScanOptions own its fields
- hide scan key generation in TryFrom
- delete a mess of glue cruft that the previous complexity necessitated
- interface to scan::scan() stays bytes

@arv this has breaking changes for js

Progress towards #218


arv (commented):
LGTM

PR #254, opened 2020-11-04
pass secondary index in scan results

phritz:
Progress towards #218


phritz:
@arv another interface change. hopefully last one. we now pass (key, secondary_key, value) to the scan callback. secondary_key will be null if it is a regular index scan, otherwise a utf8 string like key.

Issue #257, opened 2020-11-05
figure out what to do about create index transactions


phritz:
background:
https://github.com/rocicorp/replicache-sdk-js/pull/168#pullrequestreview-518184528
https://github.com/rocicorp/replicache-sdk-js/pull/168#issuecomment-718967142

right now we have the ability to create indexes within normal transactions. because creating an index can be mixed in with other kinds of db ops, we have to have a mutator for it and it must be sent to the data layer. IF we decided that create/dropindex CANNOT be mixed in with other kinds of db ops -- that if you do a create/dropindex in a tx you can't do anything else -- then index changes can go into a separate kind of commit that we dont push to the server. this is cleaner for the customers data layer, but less nice as far as the transaction api is concerned because it makes index txs special which aaron has called out as annoying.

we need to make the call about whether to keep status quo or to make index txs exclusive of puts/dels/etc. i don't really have an opinion, happy to do what @arv or @aboodman think.

aboodman:
On further thought, I think that it probably does not actually make sense to do puts/dels atomically with indexes in Replicache. This is different than a normal database where you'd be migrating old data because in Repliache the client is not the authority over the shape of the client view, the server is.

So I think we pull create/delIndex out of write tx and have it on its own for now.

phritz:
sounds right, will do. 

looks like the cleanest way to do this would be to introduce a new type of db transaction, say IndexChange, which would get the index methods from Write.

PR #260, opened 2020-11-07
make a separate transaction type for index changes

phritz:
two separate logical commits. 

@arv this is a breaking change. you now have to open index transactions with openindextransaction. they are closed the same as normal transactions with committransaction. now you can no longer do index changes in regular transactions or put/dels in index transactions. i have separated index change commits into their own commit type separate from local commits. this means that A) index changes are not returned for replay by maybeendsync and B) the customer does not have to have data layer mutators for index changes. yay!

progress towards #218 

closes https://github.com/rocicorp/repc/issues/257

### DIFF
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -49,7 +49,7 @@ impl Commit {
         mutator_args_json: &[u8],
         original_hash: Option<&str>,
         value_hash: &str,
-        indexes: &[IndexMeta],
+        indexes: &[IndexRecord],
     ) -> Commit {
         let mut builder = FlatBufferBuilder::default();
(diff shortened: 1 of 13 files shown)


src/db/commit.rs
```diff
@@ -396,22 +458,30 @@ impl<'a> Meta<'a> {
     }
 
     pub fn is_snapshot(&self) -> bool {
-        match self.typed() {
-            MetaTyped::Local(_) => false,
-            MetaTyped::Snapshot(_) => true,
-        }
+        matches!(self.typed(), MetaTyped::Snapshot(_))
```
arv, line 461: TIL

arv (commented):
LGTM

Issue #259, opened 2020-11-06
ScanOptions probably needs to be able to take the secondary key for the start key as well


aboodman:
Imagine that you are building some paginated UI from an index. You want to be able to scan starting at the last record you have seen.

Over indexes, records are uniquely identified by both their secondary and primary keys. Therefore ScanOptions.startKey needs to be able to accept both.

Low priority at the moment, but this will be an annoyance for Cron.

aboodman:
@phritz - sorry I didn't think of this sooner.

aboodman:
This would basically make the input API to scan symmetrical with the output API (which sends both secondary and primary keys), so makes sense to me.

PR #261, opened 2020-11-09
scan clean ups and enable scanning indexes by primary key

phritz:
separate logical commits

@arv this is an interface change. 
1.  start_key_exclusive => start_exclusive
2. for index scans we previously passed the secondary key value we are scanning for in start_key. however, we want to enable scanning indexes by both primary and secondary key values. so now you need to pass the secondary key value to scan for in start_secondary_key and the primary key value if any in start_key. note that for index scans the primary key (start_key) is optional and should be omitted entirely from the serialized form if not used. see comment in last commit but the difference is that in index scans specifying start_key forces an exact match on start_secondary_key. if this is too confusing then we can add a separate field.

progress towards #218 

closes https://github.com/rocicorp/repc/issues/259

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -213,7 +213,15 @@ async fn execute<'a, 'b>(
         "has" => return to_js(do_has(txn.read().await.as_read(), from_js(data)?).await),
         "get" => return to_js(do_get(txn.read().await.as_read(), from_js(data)?).await),
         "scan" => {
-            return to_js(do_scan(txn.read().await.as_read(), from_js(data.clone())?, data).await)
+            return to_js(
+                do_scan(
+                    txn.read().await.as_read(),
+                    from_js(data.clone())?,
+                    data,
+                    lc.clone(),
+                )
+                .await,
+            )
         }
         _ => (),
     }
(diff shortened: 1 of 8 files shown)


src/embed/connection.rs
```diff
@@ -505,48 +513,41 @@ async fn do_scan(
...
-            // TODO the unwrap()s below are really unfortunate
-            let (secondary, primary) = db::index::decode_index_key(pe.key).unwrap();
-            primary_key = JsValue::from_str(std::str::from_utf8(&primary).unwrap());
-            secondary_key = JsValue::from_str(&secondary);
-        } else {
-            primary_key = JsValue::from_str(std::str::from_utf8(pe.key).unwrap());
-            secondary_key = JsValue::null();
+    read.scan(req.opts, |sr: db::ScanResult<'_>| {
+        match sr {
+            db::ScanResult::Error(e) => error!(lc, "Error returning scan result: {:?}", e),
```
arv, line 525: Is this throwing in JS? I think it must. Just logging the error is going to result in a JS array of empty entries/items.

arv: Strawman proposal. Pass in a thrower function that gets called with a string.

phritz: I'm not sure why it would throw in js? error!() should just log an error. the scan will continue.

phritz: to be clear what happens is that when we get a result error instead of an item we log an error, it does not invoke the callback, and it continues on to the next item.

arv: My point is that we should throw, or tell the bindings about the error. Just logging the error leads to invalid state.

PR #266, opened 2020-11-13
implement syncid

### LINKED ISSUE #86: generate a unique syncid for every sync and include it in pull/push requests
here' how we do it in go: https://github.com/rocicorp/replicache-client/blob/c3e422ab432afcbce1ab31a33e0021d47a192b42/db/db.go#L241. it is not important to generate the unique bits the same way as in go or to have the same format EXCEPT that:
- it should include a number that is easy to see that allows us to find the next or previous sync from this one
- ideally include client id in it so in the logs it is easy to tie them together

please find the places where syncid is used and thread it through there, eg https://github.com/rocicorp/repc/blob/3678fa5b5fbcce0ed061955ebddafa790565017c/src/embed/sync.rs#L95. (yes we should prolly move this into the request payload not a header, but that's how we have it now. feel free to file an issue about that.)

to do this right we need #45 done.

syncid should be included in the context log https://github.com/rocicorp/repc/issues/85

phritz:
(turned out i didn't to move client id, but it belongs there anyway as it is a part of the sync protocol)

closes #86

### DIFF
--- a/src/sync/sync_id.rs
+++ b/src/sync/sync_id.rs
@@ -1,0 +1,58 @@
+use crate::util::uuid;
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Mutex;
+use str_macro::str;
+
+lazy_static! {
+    static ref SESSION_ID: String = {
+        let mut randoms: [u8; 4] = [0; 4];
+        uuid::make_random_numbers(&mut randoms);
+        format!(
+            "{:x}{:x}{:x}{:x}",
+            randoms[0], randoms[1], randoms[2], randoms[3]
+        )
+    };
+    static ref SYNC_COUNTERS: Mutex<HashMap<String, AtomicU32>> = Mutex::new(HashMap::new());
+}
+
+// new() returns a new syncid of the form <clientid>-<sessionid>-<sync count>.
+// The sync count enables one to find the sync following or preceeding a given
+// sync. The sessionid scopes the sync count, ensuring the syncid is
+// probabilistically unique across restarts (which is good enough).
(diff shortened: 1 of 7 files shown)


src/sync/sync_id.rs
```diff
@@ -0,0 +1,58 @@
...
+            error!("", "error getting sync count: {:?}", e);
+            return str!("busted-sync-counter");
+        }
+        Ok(m) => m,
+    };
+    let counter = match map.get(client_id) {
+        None => {
+            let counter = AtomicU32::new(0);
+            map.insert(client_id.into(), counter);
+            map.get(client_id).unwrap()
```
arv, line 35: Do we really need to insert and get?

src/sync/sync_id.rs
```diff
@@ -0,0 +1,58 @@
...
+lazy_static! {
+    static ref SESSION_ID: String = {
+        let mut randoms: [u8; 4] = [0; 4];
+        uuid::make_random_numbers(&mut randoms);
+        format!(
+            "{:x}{:x}{:x}{:x}",
+            randoms[0], randoms[1], randoms[2], randoms[3]
+        )
+    };
+    static ref SYNC_COUNTERS: Mutex<HashMap<String, AtomicU32>> = Mutex::new(HashMap::new());
```
arv, line 16: Not clear to me why we need both a Mutex and an atomic?

arv (commented):
LGTM

src/sync/sync_id.rs
phritz: yes. it gets moved into the map in 34 so we need to get it back out in 35.

PR #267, opened 2020-11-13
Make make_random_numbers and uuid return a Result

### LINKED ISSUE #248: Return result in uuid/mod
Instead of crashing

https://github.com/rocicorp/repc/blob/ba43e47b33e5e3d32df82823d512fcc5ec831d1b/src/util/uuid/mod.rs#L15

arv:
Fixes #248

### DIFF
--- a/src/sync/sync_id.rs
+++ b/src/sync/sync_id.rs
@@ -7,7 +7,7 @@ use str_macro::str;
 lazy_static! {
     static ref SESSION_ID: String = {
         let mut randoms: [u8; 4] = [0; 4];
-        uuid::make_random_numbers(&mut randoms).unwrap();
+        let _ = uuid::make_random_numbers(&mut randoms);
         format!(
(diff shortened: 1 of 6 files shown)


src/sync/sync_id.rs
```diff
@@ -7,7 +7,7 @@ use str_macro::str;
 lazy_static! {
     static ref SESSION_ID: String = {
         let mut randoms: [u8; 4] = [0; 4];
-        uuid::make_random_numbers(&mut randoms);
+        uuid::make_random_numbers(&mut randoms).unwrap();
```
arv, line 10: We could make `SESSION_ID` a `Result` if we want to be really careful.

phritz: i'd just as well ignore the result here and let it default to 0000
```
        let _ = uuid::make_random_numbers(&mut randoms);
```

phritz (approved):
LGTM

PR #268, opened 2020-11-13
Remove local_create_date from flat buffer

### LINKED ISSUE #249: Remove local_create_date in FB
https://github.com/rocicorp/repc/pull/245#pullrequestreview-519986193

We are still toddlers...

arv:
Fixes #249

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 repc.zip
 
+.DS_Store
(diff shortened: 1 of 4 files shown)


phritz (approved):
The rs files are not automatically generated so for this change to have effect you have to also run tool/flatc.sh to generate the rs file.

arv:
oops

Issue #272, opened 2020-11-15
Unexpected TimeTravel in SongBook Studio


aboodman:
@elsigh got stuck on his Replicache integration because push and pull were failing with auth errors, but the `getDataLayerAuth` callback was also not getting called.

I looked at the logs and saw "TimeTravelProhibited" in the console (sadly I don't have a screencap of this -- @elsigh ?)

When we looked at the Pull response we saw that the ClientView response was HTTP 401, and that was correctly getting sent down to client, and the lastMutationID was 0.

But the client had synced before so its base lastMutationID was > 0, thus time travel, and therefore sync was failing before it got to reauthing.

It seems like the only way this can happen is if either the client ID changes on the client somehow, or if the client data gets dropped on the server. It's a mystery how either of these could occur.

phritz:
>But the client had synced before so its base lastMutationID was > 0, thus time travel, and therefore sync was failing before it got to reauthing.
> It seems like the only way this can happen is if either the client ID changes on the client somehow, or if the client data gets dropped on the server. It's a mystery how either of these could occur.

Possibly more likely, or one of the above in combination with: the client has fresher data than the diffserver. Client sends pull request with lastmutationid N and stateid S. Diffserver does clientview request, gets a 401. Diffserver is now going to try to send the client the best diff it has: https://github.com/rocicorp/diff-server/blob/6528d96191aa0f633e2b4df3e8fd9afd78f96702/serve/pull.go#L99. If it can't find the client's state, it does a full sync: https://github.com/rocicorp/diff-server/blob/6528d96191aa0f633e2b4df3e8fd9afd78f96702/db/diff.go#L59. But... if the diffserver's head for this client is empty, it is going to try to sync back to mutationid 0 and the initial server state. Doh. The client rightfully disallows going backwards in time from N to 0 and beginsync errors. When it errors the fact that the client view returned 401 is lost. (This last part is a classic example of why errors should be signaled *within* the rpc response, rather than outside of it. More grist for the https://github.com/rocicorp/repc/issues/119 mill.) 

[...]

PR #273, opened 2020-11-17
implement setLogLevel

### LINKED ISSUE #243: implement setloglevel dispatch rpc :)
right now it is set statically https://github.com/rocicorp/repc/blob/560ec4b7d318264400c8c45f0961026bedf20b88/src/wasm.rs#L28

phritz:
@arv can you wire up the js side of this?

progress towards #272 
closes #243


aboodman (approved):

aboodman:
@arv I think that the API should be expressed as a static on Replicache or a free function since it's not specific to an instance of Replicache.

aboodman:
Also maybe we should call it `setRepcLogLevel` or some such since it doesn't affect the log level of the bindings. (And I don't really think we should make it affect the log level of the bindings, since it's pretty nice to just let the UI in the web console do that).

arv (commented):
LGTM

Issue #240, opened 2020-10-27
Verify that we aren't accidentally serializing all transactions 😬


aboodman:
https://github.com/rocicorp/repc/pull/203/files/0c7353a5f4f98d30049afd3af5d4c335f14e03f2..c27b20550acf043e9b1dd14ebc3754613eba0f8f#diff-538fa79214eb88d59c2af080b7f78d197b0c40f287f2b4b0f0114e4cbb91b9f7

phritz:
we are not serializing transactions 

we had a test that verified sequential execution for write transactions, which i improved and it still passes. we did not have a test for parallel execution for read transactions so I added one and it passes with no code changes. both of these in #270 . so looks like we are good on this front.

there are however a few things i want to understand better before i close this issue out, including at least three levels of locking in connection.rs. will post what i find here when i pick it back up next week.

phritz:
Before closing this issue out I wanted to rationalize the locking in connection. I read the code pretty closely and here's what I think I have learned. In the below by "operation" or "op" I mean an rpc called within a transaction, eg put, get, has, etc.

Desired properties:
  - A. if a write tx is "open" it is the only open tx.
    - => no two writes can be open at the same time
    - => no read and write at the same time
    - => many read tx can be open at the same time
    - by "open" we mean holding exclusive write or shared read lock on the dag store. In the current code this maps to whether the openTransaction rpc has returned: if it has, the transaction is open. (One could imagine a different implementation where openTransaction returns without the required lock, and the lock is acquired asynchronously or at first op, but that's not how the code currently works.)
  - B. **within** a write transaction a write op is exclusive of all other ops
    - => no concurrent write ops
    - => no concurrent read + write ops 
    - => read ops can (and should) execute concurrently
  - C. **within** a read transaction ops can execute concurrently

In connection there are three levels of locking in play. The transaction map looks like this:
```
enum Transaction<'a> {
    Read(db::OwnedRead<'a>),
    Write(db::Write<'a>),
}
TxnMap<'a> = RwLock<HashMap<u32, RwLock<Transaction<'a>>>>;
```
The lets call the outer lock around the hashmap the "map lock". Then we have "transaction locks" around individual Transactions in the map. And a "db lock" sits inside each Transaction. 

The things I wanted to understand are whether all these locks are necessary (i think: yes), what desired properties listed above that each provides (see below), and whether there are any unintended or less than ideal consequences from how they work (maybe).

[...]

PR #275, opened 2020-11-18
Idb transactions around operations

arv:
We now create individual IDB transactions for get/has instead of opening the IDB transaction in the kv open transaction

- [x] Verify scan
- [x] Performance tests for get/has

Towards https://github.com/rocicorp/replicache-sdk-js/issues/143

### DIFF
--- a/src/kv/idbstore.rs
+++ b/src/kv/idbstore.rs
@@ -234,7 +234,6 @@ enum WriteState {
 struct WriteTransaction<'a> {
     db: RwLockWriteGuard<'a, IdbDatabase>,
     pending: Mutex<HashMap<String, Option<Vec<u8>>>>,
-    pair: Arc<(Mutex<WriteState>, Condvar)>,
     lc: LogContext,
 }
 
@@ -242,7 +241,6 @@ impl WriteTransaction<'_> {
     fn new(db: RwLockWriteGuard<'_, IdbDatabase>, lc: LogContext) -> Result<WriteTransaction> {
         let wt = WriteTransaction {
             db,
-            pair: Arc::new((Mutex::new(WriteState::Open), Condvar::new())),
             pending: Mutex::new(HashMap::new()),
             lc,
         };
@@ -250,8 +248,10 @@ impl WriteTransaction<'_> {
         Ok(wt)
     }
 
-    fn tx_callback(&self, new_state: WriteState) -> Closure<dyn FnMut()> {
-        let pair = self.pair.clone();
+    fn tx_callback(
+        pair: Arc<(Mutex<WriteState>, Condvar)>,
+        new_state: WriteState,
+    ) -> Closure<dyn FnMut()> {
         Closure::once(move || {
             task::block_on(async move {
                 let (lock, cv) = &*pair;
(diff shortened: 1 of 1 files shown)


arv:
@phritz This might impact #240

arv:
Perf tests `master`:

```
Running benchmarks please wait...
populate 1024x1000 (clean, indexes: 0) x 3.83 MB/s ±0.0% (7 runs sampled)
populate 1024x1000 (dirty, indexes: 0) x 4.48 MB/s ±0.0% (10 runs sampled)
populate 1024x1000 (clean, indexes: 1) x 2.13 MB/s ±0.0% (5 runs sampled)
populate 1024x1000 (clean, indexes: 2) x 1.49 MB/s ±0.0% (5 runs sampled)
read tx 1024x1000 x 7.45 MB/s ±0.0% (15 runs sampled)
read tx 1024x5000 x 7.94 MB/s ±0.0% (5 runs sampled)
scan 1024x1000 x 18.78 MB/s ±0.0% (31 runs sampled)
scan 1024x5000 x 22.30 MB/s ±0.0% (9 runs sampled)
write single byte x 8.70 tx/s ±0.0% (18 runs sampled)
Done!
```

Perf tests with `arv:idb-tx-around-ops`:

```
Running benchmarks please wait...
populate 1024x1000 (clean, indexes: 0) x 3.23 MB/s ±0.0% (6 runs sampled)
populate 1024x1000 (dirty, indexes: 0) x 4.17 MB/s ±0.0% (9 runs sampled)
populate 1024x1000 (clean, indexes: 1) x 1.99 MB/s ±0.0% (5 runs sampled)
populate 1024x1000 (clean, indexes: 2) x 1.38 MB/s ±0.0% (5 runs sampled)
read tx 1024x1000 x 7.23 MB/s ±0.0% (15 runs sampled)
read tx 1024x5000 x 8.04 MB/s ±0.0% (5 runs sampled)
scan 1024x1000 x 16.01 MB/s ±0.0% (32 runs sampled)
scan 1024x5000 x 23.36 MB/s ±0.0% (10 runs sampled)
write single byte x 5.78 tx/s ±0.0% (12 runs sampled)
Done!
```

arv:
It seems that opening an IDB read transaction has no discernible performance overhead. That is good news because it means that we do not need to refactor scan to open the IDB tx before we scan all the items.

Ready for review!

phritz (approved):
> Make IdbStore use a transaction for every get/has/scan

This seems like a good idea I guess but can you say *why* we want to do this?

The model for idb is now different than for higher levels in that the transaction isn't opened until an operation is started, as opposed to dg eb where creating the transaction opens it. I will make a note in #240 that one of us should look more closely at this level (or, if you already understand it and how it interacts with higher levels, feel free to drop your knowledge into 240).

aboodman:
The writes seem a little slower to me:

master:

populate 1024x1000 (clean, indexes: 0) x 3.83 MB/s ±0.0% (7 runs sampled)
populate 1024x1000 (dirty, indexes: 0) x 4.48 MB/s ±0.0% (10 runs sampled)
populate 1024x1000 (clean, indexes: 1) x 2.13 MB/s ±0.0% (5 runs sampled)
populate 1024x1000 (clean, indexes: 2) x 1.49 MB/s ±0.0% (5 runs sampled)

write single byte x 8.70 tx/s ±0.0% (18 runs sampled)

branch:

populate 1024x1000 (clean, indexes: 0) x 3.23 MB/s ±0.0% (6 runs sampled)
populate 1024x1000 (dirty, indexes: 0) x 4.17 MB/s ±0.0% (9 runs sampled)
populate 1024x1000 (clean, indexes: 1) x 1.99 MB/s ±0.0% (5 runs sampled)
populate 1024x1000 (clean, indexes: 2) x 1.38 MB/s ±0.0% (5 runs sampled)

write single byte x 5.78 tx/s ±0.0% (12 runs sampled)

Note these are medians on a single machine, so probably reliable. You
could do it a few more times to be sure.

This drop seems reasonable to me, since we're now doing more work.
Frankly better than I expected. So yay, I'm ok with this to proceed w/
serviceworker experiments.

The one exception is the write-single-byte test, which got a lot
slower. This test does:

```

  const rep = await makeRep();
  const write = rep.register('write', tx => tx.put('k', i % 10));
  bench.name = 'write single byte';
  bench.size = 1;
  bench.formatter = formatTxPerSecond;
  bench.reset();

  await write(null);

  bench.stop();
  await rep.close();

```

So it seems it is doing the exact same amt of work in both master and
branch - open a transaction, writing a byte, then committing. Just
doing it a little later. Surprised to see any difference in that one.
Worth investigating.

Erik, when I was investigating this stuff I found the timeline view in
Chrome invaluable. If you run with what we currently call a "debug"
build, you'll get symbols you can see in timeline, and you can easily
see where time is getting spent inside Rust.

arv:
I'm also surprised that write changed. Logically nothing should have changed there.

arv:
Not doing this because we are not doing SW so it is not needed.

PR #279, opened 2020-11-30
Revert "Remove local_create_date from flat buffer (#268)"

phritz:
This reverts commit e7e33b4902ea0fea3cd11be76c7cf4e933920138.

Reverting because this is a breaking change and we have a customer in production.

Re-opens #249


arv:
LGTM

PR #289, opened 2021-02-11
For discussion: Update BSL.txt to enable non-commercial uses and bump number of allowed commercial clients

aboodman:
Note this requires specifically restricting creation of competing products.

### DIFF
--- a/licenses/BSL.txt
+++ b/licenses/BSL.txt
@@ -7,9 +7,15 @@ Licensor:             Rocicorp, LLC
 Licensed Work:        repc 0.18.0
                       The Licensed Work is (c) 2020 Rocicorp, LLC.
 
-Additional Use Grant: You may use the Licensed Work when your application uses
-                      the Licensed Work with a total of less than 100 client
-                      instances for any purpose.
+Additional Use Grant: You may make use of the Licensed Work, provided you
+                      do not use it in any of the following ways:
+
+                      * Create a synchronization engine. A "synchronization
+                     engine" is a commercial or non-commercial offering that
+                     synchronizes data between two or more computers.
+
+                     * Use it for commercial purposes with more than 500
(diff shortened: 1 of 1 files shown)


aboodman:
Inspired by:

https://github.com/zerotier/ZeroTierOne/blob/master/LICENSE.txt and
https://github.com/cockroachdb/cockroach/blob/master/licenses/BSL.txt

licenses/BSL.txt
```diff
@@ -7,9 +7,15 @@ Licensor:             Rocicorp, LLC
...
 
-Additional Use Grant: You may use the Licensed Work when your application uses
-                      the Licensed Work with a total of less than 100 client
-                      instances for any purpose.
+Additional Use Grant: You may make use of the Licensed Work, provided you
+                      do not use it in any of the following ways:
+
+                      * Create a synchronization engine. A "synchronization
+                     engine" is a commercial or non-commercial offering that
+                     synchronizes data between two or more computers.
```
arv, line 15: Not sure if this is clear enough but I'm not sure how to improve it.

Maybe we can say something like "no derivative work" -- https://creativecommons.org/characteristic/nd?lang=en

arv:
LGTM

licenses/BSL.txt
```diff
@@ -7,9 +7,15 @@ Licensor:             Rocicorp, LLC
...
+
+                     * Use it for commercial purposes with more than 500
+                     monthly active client instances.
```
phritz, line 18: I think we should also say. `... In order to use it for commercial purposes with more than 500 monthly active client instances, a license is required: https://replicache.dev/#price.`

Also guess we should have a link right there in the pricing info to email us for a license.

phritz (approved):

Issue #290, opened 2021-02-12
eliminate diffserver and move to diffsync


phritz:
Proposal: https://www.notion.so/Differential-Client-View-12be3a636c9f404b88d49ecbd100a694

TODOs in repc:
- remove checksum from pull and checksum itself. 
  - it is a question if we do this in a breaking vs nonbreaking way (eg, remove checksum from commit vs not). think it makes sense to do this in a breaking way bc there are commit changes beyond checksum (storing cookie for example). which precipitates the question of what to do with existing customers and how to manage this kind of change in the future.
- split BeginSync into separate push and pull calls. Call them TryPush and TryPull. Move pull from mod.rs into its own source file. it's a good question as to whether to do this work before the next item, or as part of it, whoever is doing the work can see which seems like less work.
- Update TryPush which was formerly part of [BeginSyncRequest](https://github.com/rocicorp/repc/blob/15c24420890ff639ba7bd2f1015d4b23c8cdae0b/src/sync/types.rs#L5) to have: pushurl, pushurlauth. Evaluate whether the response should have syncinfo/batchpushinfo.
- Update TryPull formerly part of [BeginSyncRequest](https://github.com/rocicorp/repc/blob/15c24420890ff639ba7bd2f1015d4b23c8cdae0b/src/sync/types.rs#L5) to have: pullurl, pullurlauth. Evaluate whether the response should have syncinfo/clientviewinfo suitably renamed; if so repc will now have to add it instead of copy it from diffs response.
- rename [batchpushrequest](https://github.com/rocicorp/repc/blob/15c24420890ff639ba7bd2f1015d4b23c8cdae0b/src/sync/push.rs#L8)/respone => pushrequest/response. 
- replace the existing pullrequest/response and clientviewrequest/response with a new pullrequest. This pullrequest should have at minimum: clientid, last_mutation_id, cookie, pullversion (formerly version). Pullresponse should have cookie, lastmutationid, patch. Remove the concept of stateid everywhere (or replace it with cookie). 
- Change patch value to be JSON, not a [string](https://github.com/rocicorp/repc/blob/15c24420890ff639ba7bd2f1015d4b23c8cdae0b/src/sync/patch.rs#L16) because we no longer care about checksumming.
- add a pushversion to pushrequest so we can change it when we need to (we have already used the pullversion extensively)
- add an appversion to trypush and trypull and include it in the pull and push requests so they can version their client view
[...]

PR #296, opened 2021-02-23
Rename Pull/Push to TryPull/TryPush

arv:
Towards #290

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -189,9 +189,9 @@ async fn execute<'a, 'b>(
         "closeTransaction" => return to_js(do_close(ctx, from_js(data)?).await),
         "setLogLevel" => return to_js(do_set_log_level(ctx, from_js(data)?).await),
 
-        "push" => return to_js(do_try_push(ctx, from_js(data)?).await),
-        "beginPull" => return to_js(do_try_begin_pull(ctx, from_js(data)?).await),
-        "maybeEndPull" => return to_js(do_maybe_end_try_pull(ctx, from_js(data)?).await),
+        "tryPush" => return to_js(do_try_push(ctx, from_js(data)?).await),
+        "beginTryPull" => return to_js(do_begin_try_pull(ctx, from_js(data)?).await),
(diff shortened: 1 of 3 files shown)


src/embed/connection.rs
```diff
@@ -189,9 +189,9 @@ async fn execute<'a, 'b>(
         "closeTransaction" => return to_js(do_close(ctx, from_js(data)?).await),
         "setLogLevel" => return to_js(do_set_log_level(ctx, from_js(data)?).await),
 
-        "push" => return to_js(do_push(ctx, from_js(data)?).await),
-        "beginPull" => return to_js(do_begin_pull(ctx, from_js(data)?).await),
-        "maybeEndPull" => return to_js(do_maybe_end_pull(ctx, from_js(data)?).await),
+        "push" => return to_js(do_try_push(ctx, from_js(data)?).await),
```
arv, line 192: Not clear if we also want "try" in the RPC? I think yes for consistency.

phritz: yes agree

src/embed/connection.rs
```diff
@@ -671,13 +671,13 @@ async fn do_push<'a, 'b>(
 
     let batch_push_info =
         sync::push(&sync_id, ctx.store, ctx.lc, ctx.client_id, &pusher, req).await?;
-    Ok(sync::PushResponse { batch_push_info })
+    Ok(sync::TryPushResponse { batch_push_info })
 }
 
-async fn do_begin_pull<'a, 'b>(
+async fn do_try_begin_pull<'a, 'b>(
```
phritz, line 677: the try is in two different places in the name for pull: do_maybe_end_TRY_pull and do_TRY_begin_pull. since i think the name of the thing is "try pull" it should be do_begin_try_pull and do_maybe_end_try_pull. i realize it is a mouthful.

phritz (approved):

src/embed/connection.rs
arv: ha!, I was going to make it TryX but I failed to consistently rename this

PR #297, opened 2021-02-24
Move push to push.rs

arv:
Towards #290

### DIFF
--- a/src/embed/connection.rs
+++ b/src/embed/connection.rs
@@ -1,12 +1,12 @@
 use super::dispatch::Request;
 use super::types::*;
 use crate::dag;
-use crate::db;
 use crate::fetch;
 use crate::sync;
 use crate::util::rlog;
 use crate::util::rlog::LogContext;
 use crate::util::to_debug;
+use crate::{db, sync::push::push};
 use async_std::stream::StreamExt;
 use async_std::sync::{Receiver, RecvError, RwLock};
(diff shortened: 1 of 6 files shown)


src/embed/connection.rs
```diff
@@ -1,12 +1,12 @@
 use super::dispatch::Request;
 use super::types::*;
 use crate::dag;
-use crate::db;
 use crate::fetch;
 use crate::sync;
 use crate::util::rlog;
 use crate::util::rlog::LogContext;
 use crate::util::to_debug;
+use crate::{db, sync::push::push};
```
phritz, line 9: so weird why it would join these two but not the others above (fetch, sync, etc).

src/sync/push.rs
```diff
@@ -116,6 +119,79 @@ pub enum PushError {
     SerializePushError(serde_json::error::Error),
 }
 
+/// Find pending commits between the base snapshot and the main head and push
```
phritz, line 122: in the last pr this comment should have gone inside the function, not on it. maybe move it to above line 134 (let dag_read =) below?

phritz (approved):

src/embed/connection.rs
arv: Yeah. I don't pay attention to these lines... Moving things around manually before fmt makes it more consistent.

PR #302, opened 2021-02-25
Move test_push from mod.rs to push.rs

arv:
Towards #290


arv:
TBR

src/sync/push.rs
```diff
@@ -195,9 +194,19 @@ pub async fn push(
 #[cfg(not(target_arch = "wasm32"))]
```
phritz, line 194: same comment here re arch. only the test that uses the network needs this target (test_push). should probably rename also to avoid collision something like test_try_push_http or similar.

phritz (commented):
LGTM

PR #304, opened 2021-02-26
Move pull test part into pull

arv:
Towards #290


arv:
TBR

src/sync/pull.rs
```diff
@@ -367,11 +368,22 @@ pub enum PullError {
 #[cfg(not(target_arch = "wasm32"))]
```
phritz, line 368: Hmm. We need the not wasm32 directive around any test that uses rust networking. The only test here that does that is test_pull_http_part. The test_begin/end_try_pull test does not use http, they use a fakepuller. I think the right thing to do here is to have two separate test modules: tests, which is tests that run regardless of architecture and includes the test_begin/end_try_pull, and then rust_tests, which run only on not-wasm32 and includes test_pull_http_part.

src/sync/pull.rs
```diff
@@ -490,4 +502,628 @@ mod tests {
...
+                None => {
+                    let r = self.resp.as_ref();
+                    Ok(r.unwrap().clone())
+                }
+            }
+        }
+    }
+
+    #[async_std::test]
+    async fn test_maybe_end_pull() {
```
phritz, line 972: think this should be test_maybe_end_try_pull and the function should be maybe_end_try_pull.

phritz (commented):
LGTM

src/sync/pull.rs
arv: Can't have multiple test modules but I moved the attribute to function and use statements.

PR #305, opened 2021-02-26
Replace {BatchPush,ClientView}Info with HttpRequestInfo

arv:
Towards #290

### DIFF
--- a/src/sync/types.rs
+++ b/src/sync/types.rs
@@ -1,22 +1,12 @@
 use serde::{Deserialize, Serialize};
-use std::default::Default;
 
 use crate::{dag, db, util::rlog};
 
 use super::{patch, PullError};
 
-#[derive(Debug, Deserialize, Serialize)]
-#[cfg_attr(test, derive(PartialEq))]
-pub struct BatchPushInfo {
-    #[serde(rename = "httpStatusCode")]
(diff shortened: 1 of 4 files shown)


src/sync/types.rs
```diff
@@ -1,22 +1,12 @@
...
-    pub http_status_code: u16,
-    #[serde(rename = "errorMessage")]
-    pub error_message: String,
-    // TODO BatchPushResponse BatchPushResponse `json:"batchPushResponse"`
-}
-
-#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct ClientViewInfo {
+#[derive(Serialize, Deserialize)]
+#[cfg_attr(test, derive(Clone, Debug, PartialEq))]
```
arv, line 8: Not sure if is worth doing the cfg_attr dance? Seems slightly cleaner to have less derives.

arv:
TBR

src/sync/types.rs
phritz: It doesn't hurt, we should probably be more careful with this kind of thing in general. Suspect as aaron has said there will come a point in time when we find market fit that we do a big push to reduce size.

phritz (commented):
LGTM

PR #311, opened 2021-02-26
Add push version

arv:
Towards #290

### DIFF
--- a/src/sync/pull.rs
+++ b/src/sync/pull.rs
@@ -16,6 +16,8 @@ use std::collections::hash_map::HashMap;
 use std::default::Default;
 use std::fmt::Debug;
 
+const PULL_VERSION: u32 = 0;
+
 pub async fn begin_pull(
     client_id: String,
     begin_pull_req: BeginTryPullRequest,
@@ -51,7 +53,7 @@ pub async fn begin_pull(
         client_id,
         cookie: base_cookie.clone(),
         last_mutation_id: base_snapshot.mutation_id(),
-        pull_version: 0,
+        pull_version: PULL_VERSION,
     };
     debug!(lc, "Starting pull...");
     let pull_timer = rlog::Timer::new().map_err(InternalTimerError)?;
@@ -255,6 +257,7 @@ pub struct PullRequest {
     pub cookie: String,
     #[serde(rename = "lastMutationID")]
     pub last_mutation_id: u64,
+    #[serde(rename = "pullVersion")]
     pub pull_version: u32,
 }
 
@@ -378,7 +381,7 @@ mod tests {
                 client_id: str!("client_id"),
                 cookie: str!("cookie"),
                 last_mutation_id: 123,
-                pull_version: 0,
+                pull_version: PULL_VERSION,
             };
             // EXP_BODY must be 'static to be used in HTTP handler closure.
             static ref EXP_BODY: String = serde_json::to_string(&*PULL_REQ).unwrap();
@@ -564,7 +567,7 @@ mod tests {
             client_id: client_id.clone(),
             cookie: base_cookie.clone(),
             last_mutation_id: base_last_mutation_id,
-            pull_version: 0,
+            pull_version: PULL_VERSION,
         };
 
         let cases: Vec<Case> = vec![
--- a/src/sync/push.rs
+++ b/src/sync/push.rs
@@ -6,11 +6,15 @@ use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use str_macro::str;
 
+const PUSH_VERSION: u32 = 0;
+
 #[derive(Debug, Deserialize, PartialEq, Serialize)]
 pub struct PushRequest {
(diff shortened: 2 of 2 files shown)


arv:
Doesn't really do much 🤷

src/sync/pull.rs
```diff
@@ -255,6 +257,7 @@ pub struct PullRequest {
     pub cookie: String,
     #[serde(rename = "lastMutationID")]
     pub last_mutation_id: u64,
+    #[serde(rename = "pullVersion")]
```
arv, line 260: For consistency but we should rename all of them to snake_case. I added an entry to the task list.

https://www.w3.org/TR/design-principles/#casing-rules

arv:
TBR

src/sync/pull.rs
aboodman: Is this really followed? Seems icky at least for js consumers because then the resulting js objects conflict with js style.

I personally prefer it as is but if you can show me that everyone else is using snake case I’ll go along.

arv: The JS API is still all camelCase. This is just for the json.

https://developer.mozilla.org/en-US/docs/Web/Manifest
https://developer.chrome.com/docs/extensions/mv3/manifest/

aboodman:
On Sat, Feb 27, 2021 at 12:43 PM Erik Arvidsson <notifications@github.com>
wrote:

> *@arv* commented on this pull request.
>
> In src/sync/pull.rs
> <https://github.com/rocicorp/repc/pull/311#discussion_r584200357>:
>
> > @@ -255,6 +257,7 @@ pub struct PullRequest {
>      pub cookie: String,
>      #[serde(rename = "lastMutationID")]
>      pub last_mutation_id: u64,
> +    #[serde(rename = "pullVersion")]
>
> The JS API is still all camelCase. This is just for the json.
>
> https://developer.mozilla.org/en-US/docs/Web/Manifest
> https://developer.chrome.com/docs/extensions/mv3/manifest/
>
What is the distinction between the JS API and the JSON?

In Next.js, I consume the JSON and use it as JS objects via JSON.parse.
There's not really a decode step that I have fine control over...

The links you provided are for web platform APIs, and I don't feel like
it's really the same. I'm wondering if there's a strong convention to
prefer snake case for web service APIs.

aboodman:
On Sat, Feb 27, 2021 at 12:58 PM Aaron Boodman <aaron@aaronboodman.com>
wrote:

>
>
> On Sat, Feb 27, 2021 at 12:43 PM Erik Arvidsson <notifications@github.com>
> wrote:
>
>> *@arv* commented on this pull request.
>> ------------------------------
>>
>> In src/sync/pull.rs
>> <https://github.com/rocicorp/repc/pull/311#discussion_r584200357>:
>>
>> > @@ -255,6 +257,7 @@ pub struct PullRequest {
>>      pub cookie: String,
>>      #[serde(rename = "lastMutationID")]
>>      pub last_mutation_id: u64,
>> +    #[serde(rename = "pullVersion")]
>>
>> The JS API is still all camelCase. This is just for the json.
>>
>> https://developer.mozilla.org/en-US/docs/Web/Manifest
>> https://developer.chrome.com/docs/extensions/mv3/manifest/
>>
> What is the distinction between the JS API and the JSON?
>
> In Next.js, I consume the JSON and use it as JS objects via JSON.parse.
> There's not really a decode step that I have fine control over...
>
> The links you provided are for web platform APIs, and I don't feel like
> it's really the same. I'm wondering if there's a strong convention to
> prefer snake case for web service APIs.
>

I see what you mean. You're saying the client side API is still camel case.
I get that -- it's only the server developer who sees these names.

It looks like there's no strong convention. The google styleguide says
camelcase:
https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format,
but google maps uses snake :). Facebook and Twitter APIs both use snakecase.

I guess since I'm currently in JS land on the server constructing these
responses, so that's where the slight preference that it feels js-y is
coming from.

Guess I don't have that strong a preference is you feel strongly about
this. I can see the argument your way.

arv:
As long as it is consistent I do not care too much either. I was just trying to match some kind of industry convention but it seems like there really isn't enough of one.

Using camelCase is nicer in some way because we do not need eslint comments everywhere.

I'm just going to stick with what we have.

src/sync/pull.rs
```diff
@@ -16,6 +16,8 @@ use std::collections::hash_map::HashMap;
 use std::default::Default;
 use std::fmt::Debug;
 
+const PULL_VERSION: u32 = 0;
```
phritz, line 19: Can you put this next to the PullRequest definition, and maybe move the comment describing version 0 to it?

src/sync/pull.rs
phritz: Strong preference for narrow gully hungarian notation, which would make this version integer `VIpULLvERSION`

phritz (commented):
LGTM

PR #313, opened 2021-03-01
Rename sync_id to request_id

arv:
pull and push already got new own IDS when I changed from sync to
pull/push.

Towards #290

### DIFF
--- a/src/sync/request_id.rs
+++ b/src/sync/request_id.rs
@@ -16,9 +16,9 @@ lazy_static! {
     static ref SYNC_COUNTERS: Mutex<HashMap<String, AtomicU32>> = Mutex::new(HashMap::new());
 }
 
-// new() returns a new syncid of the form <clientid>-<sessionid>-<sync count>.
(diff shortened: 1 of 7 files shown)


arv:
TBR

src/sync/request_id.rs
```diff
@@ -16,9 +16,9 @@ lazy_static! {
     static ref SYNC_COUNTERS: Mutex<HashMap<String, AtomicU32>> = Mutex::new(HashMap::new());
 }
 
-// new() returns a new syncid of the form <clientid>-<sessionid>-<sync count>.
+// new() returns a new request_id of the form <clientid>-<sessionid>-<sync count>.
 // The sync count enables one to find the sync following or preceeding a given
```
phritz, line 20: Here in lines 19-41 we should probably stop referring to them as "sync count" and "SYNC_COUNTER" and call it request counter.

phritz (commented):
Once section to fix up otherwise LGTM

PR #314, opened 2021-03-02
Remove http_request_info from PullResponse

arv:
We still need it in BeginTryPullResponse so we unpack the error.

Towards #290

### DIFF
--- a/src/sync/pull.rs
+++ b/src/sync/pull.rs
@@ -334,6 +334,10 @@ impl FetchPuller<'_> {
 
 #[async_trait(?Send)]
 impl Puller for FetchPuller<'_> {
+    // A failed HTTP response (non 200) is not an error. In that case we get
+    // `None` for the `PullResponse`. We get errors for a few non HTTP related
+    // reasons such as if we fail to create a Request object, the call to fetch
+    // fails or the response is not the expected JSON format.
     async fn pull(
         &self,
         pull_req: &PullRequest,


arv:
Feels a little bit ugly to use an `PullError` in `Puller::pull` and then just unpack it in the caller. An alternative could be to not treat HTTP errors as errors and have `Puller::pull` return `(Option<PullResponse>, HttpRequestInfo)`?

phritz:
> Feels a little bit ugly to use an PullError in Puller::pull and then just unpack it in the caller. An alternative could be to not treat HTTP errors as errors and have Puller::pull return (Option<PullResponse>, HttpRequestInfo)?
> ... or leave the error and start using structured errors all the way up to JS.

Yes either of those might be better. The first sounds easier. If you feel like going that route we should probably have the pusher work the same way.

src/sync/pull.rs
```diff
@@ -59,8 +60,18 @@ pub async fn begin_pull(
     let pull_timer = rlog::Timer::new().map_err(InternalTimerError)?;
     let pull_resp = puller
         .pull(&pull_req, &pull_url, &pull_auth, &request_id)
-        .await
-        .map_err(PullFailed)?;
+        .await;
+    if let Err(PullError::FetchNotOk(status_code, body)) = pull_resp {
```
phritz, line 64: Yeah, looking now at the code your suggestion to have puller return its httpinfo sounds like the way to go. The puller should pull and the caller can decide what to do with the info and response.

src/sync/pull.rs
```diff
@@ -59,8 +60,18 @@ pub async fn begin_pull(
...
+        return Ok(BeginTryPullResponse {
```
phritz, line 65: I would have thought this would still be an Err? Is it because we can't return the httpinfo in an error? Can we not have a(nother, lame) special case error that js recognizes?

src/sync/pull.rs
```diff
@@ -141,7 +152,10 @@ pub async fn begin_pull(
     let commit_hash = db_write.commit(SYNC_HEAD_NAME).await.map_err(CommitError)?;
 
     Ok(BeginTryPullResponse {
-        http_request_info: pull_resp.http_request_info,
+        http_request_info: HttpRequestInfo {
+            http_status_code: 200,
```
phritz, line 156: This makes me kind of uncomfortable. I would rather take the info from the pull...

arv: Me too. I'm going to use a tuple.

arv:
I'll do Pusher in a new PR

phritz:
> I'll do Pusher in a new PR

Whoops, ignore my recent comment about fetchpuller :) I was doing reviews in reverse chronological order doh.

src/sync/pull.rs
```diff
@@ -991,19 +1080,26 @@ mod tests {
...
 
             // Check that BeginTryPullResponse is filled like we would expect.
-            if c.exp_err.is_none() {
-                if let Ok(pull_response) = &c.pull_result {
-                    assert_eq!(
-                        pull_response.http_request_info,
-                        got_resp.unwrap().http_request_info
-                    );
-                }
+            assert_eq!(result.is_ok(), c.exp_begin_try_pull_result.is_ok());
```
phritz, line 1089: The comment is wrong :( Here we only want to check that *HttpRequestInfo* is filled like we would expect.

src/sync/pull.rs
```diff
@@ -991,19 +1080,26 @@ mod tests {
...
+            if let Ok(result) = result {
+                // Do not check syncHead... it is not stable?
```
phritz, line 1091: Synchead is checked way up above starting around line 992. The hash is not stable so up there we check all its fields. Sorry the comment above was misleading!

src/sync/pull.rs
```diff
@@ -798,15 +886,24 @@ mod tests {
...
-                exp_err: Some("TimeTravel"),
                 exp_new_sync_head: None,
+                exp_begin_try_pull_result: Err(BeginTryPullError::TimeTravelProhibited(str!(
+                    "base lastMutationID 1 is > than client view lastMutationID 0; ignoring client view"
+                ))),
             },
             Case {
                 name: "pull 500s -> beginpull errors",
                 num_pending_mutations: 0,
                 pull_result: Err(str!("FetchNotOk(500)")),
```
phritz, line 897: I guess this can be anything we want because we no longer have this specific error. "BOOM" would work just as well.

phritz (approved):
LGTM

src/sync/pull.rs
arv: Yup. This comment is dead "code". I had it before I realized it was checked earlier.

I will update comments.

src/sync/pull.rs
arv: Let me check...

PR #316, opened 2021-03-02
properly handle nop pulls

phritz:
In particular:
- no patch, same lmid, same cookie => is a nop, do not create a new synchead in begintrypull
- no patch, same lmid, NEW cookie => is NOT a nop; do create a new synchead


arv (commented):
LGTM

PR #318, opened 2021-03-02
Make Pusher always return http info

arv:


src/sync/push.rs
```diff
@@ -53,7 +53,7 @@ pub trait Pusher {
         push_url: &str,
         push_auth: &str,
         request_id: &str,
-    ) -> Result<PushResponse, PushError>;
+    ) -> Result<(Option<PushResponse>, HttpRequestInfo), PushError>;
```
phritz, line 56: Would you mind adding a comment here to the trait or fetchpusher push implementation that says under what circumstances it errors? I think it errors only if there is a problem sending the request or parsing the response on a 200. Any non-200 response is Ok with an empty pushresponse.

src/sync/push.rs
```diff
@@ -168,29 +176,16 @@ pub async fn push(
         };
         debug!(lc, "Starting push...");
         let push_timer = rlog::Timer::new().map_err(InternalTimerError)?;
-        let push_resp = pusher
+        let (_push_resp, req_info) = pusher
```
phritz, line 179: so much better, thanks for doing this!

phritz (approved):
LGTM

You could make the fetchpusher work this way too if you wanted...

PR #322, opened 2021-03-03
change cookie from string to JSON

phritz:
the cookie defaults to `null`, which is the value that will be sent with the first PullRequest, and will be the default if the cookie is missing from the PullRequest/Response. If we want to change it to "" or something else we just have to update the  `#[serde(default)]` on the field in the request and response.

### DIFF
--- a/src/db/commit.fbs
+++ b/src/db/commit.fbs
@@ -22,7 +22,7 @@ table LocalMeta {
 // Commit metadata specific to snapshot commits.
 table SnapshotMeta {
     last_mutation_id: ulong;
-    cookie: string;
+    cookie_json: [ubyte];
 }
 
 // Commit metadata specific to the type of commit.
--- a/src/db/commit.rs
+++ b/src/db/commit.rs
@@ -70,14 +70,14 @@ impl Commit {
     pub fn new_snapshot(
         basis_hash: Option<&str>,
         last_mutation_id: u64,
-        cookie: &str,
+        cookie_json: &[u8],
         value_hash: &str,
         indexes: &[IndexRecord],
     ) -> Commit {
         let mut builder = FlatBufferBuilder::default();
         let snapshot_meta_args = &commit_fb::SnapshotMetaArgs {
             last_mutation_id,
-            cookie: builder.create_string(cookie).into(),
+            cookie_json: builder.create_vector(cookie_json).into(),
         };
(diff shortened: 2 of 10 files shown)


src/db/commit.fbs
```diff
@@ -22,7 +22,7 @@ table LocalMeta {
 // Commit metadata specific to snapshot commits.
 table SnapshotMeta {
     last_mutation_id: ulong;
-    cookie: string;
+    cookie_json: [ubyte];
```
arv, line 25: I'm curious why we prefer `[ubyte]` over string?

src/db/commit.rs
```diff
@@ -391,12 +393,16 @@ impl Commit {
...
+    ) -> Result<(u64, serde_json::Value), InternalProgrammerError> {
+        use InternalProgrammerError::*;
         match c.meta().typed() {
-            MetaTyped::Snapshot(sm) => Ok((sm.last_mutation_id(), sm.cookie().to_string())),
-            _ => Err(InternalProgrammerError::WrongType(str!(
-                "Snapshot meta expected"
-            ))),
+            MetaTyped::Snapshot(sm) => Ok((
+                sm.last_mutation_id(),
+                serde_json::from_slice(sm.cookie_json()).map_err(InvalidCookieJSON)?,
```
arv, line 403: Rust wants `Json` over `JSON`

src/db/write.rs
```diff
@@ -87,7 +87,8 @@ impl<'a> Write<'a> {
     pub async fn new_snapshot(
         whence: Whence,
         last_mutation_id: u64,
-        cookie: String,
+        cookie: serde_json::Value,
+        //fritz
```
arv, line 91: ?

arv (commented):
LGTM

src/db/commit.fbs
phritz: For better or worse I was following the convention established by the args: https://github.com/rocicorp/repc/blob/962d45b714dd285628855887fa656baceb386e9c/src/db/commit.fbs#L18. I suspect since it's bytes under the hood, that's why someone chose [ubyte].

src/db/commit.rs
phritz: bah, me hates. but fixed https://github.com/rocicorp/repc/pull/323

src/db/write.rs
phritz: heh, removed https://github.com/rocicorp/repc/pull/323

PR #324, opened 2021-03-03
Use our own JSON patch format

### LINKED ISSUE #315: Let's ditch JSON Patch
As a part of #290 we accidentally (or at least non-consciously) made JSON Patch public API. It's a lot more general and powerful than we currently need, and makes using slash-separated keys painful.

Let's do something a lot narrower that only supports the features we actually have:

- clearing the entire keyspace
- deleting an entry by key
- setting a key/value pair

arv:
We are using a simpler patch format than the JSON Patch spec that has a
more limited scope.

See #315 for "spec".

Fixes #315

### DIFF
--- a/src/sync/patch.rs
+++ b/src/sync/patch.rs
@@ -1,64 +1,48 @@
 use crate::db;
 use crate::util::rlog;
 use serde::Deserialize;
-use serde_json::json;
-use std::default::Default;
 
-const OP_ADD: &str = "add";
-const OP_REMOVE: &str = "remove";
-const OP_REPLACE: &str = "replace";
-
-#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
-pub struct Operation {
-    pub op: String,
-    pub path: String,
-    #[serde(default)]
-    pub value: serde_json::Value,
+#[derive(Deserialize)]
+#[cfg_attr(test, derive(Clone, Debug, PartialEq))]
+#[serde(tag = "op")]
+pub enum Operation {
+    #[serde(rename = "put")]
+    Put {
+        key: String,
+        value: serde_json::Value,
+    },
+    #[serde(rename = "del")]
+    Del { key: String },
+    #[serde(rename = "clear")]
+    Clear,
 }
 
 pub async fn apply(db_write: &mut db::Write<'_>, patch: &[Operation]) -> Result<(), PatchError> {
     use PatchError::*;
     for op in patch.iter() {
-        // Special case `{"op": "replace", "path": "", "value": {}}`
-        // which means replace the root with a new map, in other words, clear
-        // the map.
-        if op.path.is_empty() {
-            if op.op == OP_REPLACE && op.value == json!({}) {
-                db_write.clear().await.map_err(ClearError)?;
-                continue;
-            }
-            return Err(InvalidPath(op.path.clone()));
-        }
-
-        let mut chars = op.path.chars();
-        if chars.next() != Some('/') {
-            return Err(InvalidPath(op.path.clone()));
-        }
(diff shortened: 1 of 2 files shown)


src/sync/patch.rs
```diff
@@ -1,66 +1,50 @@
...
-
-#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
-pub struct Operation {
-    pub op: String,
-    pub path: String,
-    #[serde(default)]
-    pub value: serde_json::Value,
+#[derive(Deserialize)]
+#[cfg_attr(test, derive(Clone, Debug, PartialEq))]
+#[serde(tag = "op")]
```
arv, line 7: This worked out nicely. I was half expecting to have to write my own Deserializer.

arv:
TBR

src/sync/patch.rs
phritz: yeah that is really nice, TIL

src/sync/patch.rs
```diff
@@ -1,66 +1,50 @@
...
-        }
-        let key = json_pointer_unescape(chars.as_str()).as_bytes().to_vec();
-
-        match op.op.as_str() {
-            OP_ADD | OP_REPLACE => {
-                let value = serde_json::to_vec(&op.value).map_err(InvalidValue)?;
+        match op {
+            Operation::Put { key, value } => {
+                let key = key.as_bytes().to_vec();
+                let value = serde_json::to_vec(value).map_err(InvalidValue)?;
```
phritz, line 26: @aboodman note that our policy is not error out if all there is a problem with any op. I think this is right because we don't want to partially apply an update. Just calling it out.

src/sync/patch.rs
```diff
@@ -1,66 +1,50 @@
...
                 db_write
                     .put(rlog::LogContext::new(), key, value)
                     .await
                     .map_err(PutError)?;
             }
-            // Should we error if we try to remove a key that doesn't exist?
-            OP_REMOVE => {
+            Operation::Del { key } => {
+                // Should we error if we try to remove a key that doesn't exist?
```
phritz, line 33: Answer: I think no. Let's change this comment to
```
// Note it is not an error to del a key that does not exist.
```

src/sync/patch.rs
```diff
@@ -100,96 +85,114 @@ mod tests {
...
-                ],
+                name: "put/remove empty key",
+                patch: json!([
+                    {"op": "put", "key": "", "value": "empty"},
+                    {"op": "del", "key": ""}
+                ]),
                 exp_err: None,
                 exp_map: Some(map!("key" => "value")),
             },
             // Remove once all the other layers no longer depend on this.
```
phritz, line 137: This comment should be removed.

src/sync/patch.rs
```diff
@@ -100,96 +85,114 @@ mod tests {
...
             Case {
                 name: "top-level remove",
```
phritz, line 139: case name can change to simply "clear"

phritz (commented):
Really nice.

src/sync/patch.rs
aboodman: To be clear the policy is to exit early in case of error and abort the whole patch, right? If so, lgtm.

src/sync/patch.rs
aboodman: To keep it easy to understand let's have the semantics be just like the corresponding operations on `WriteTransaction`. So in this case, no error.

src/sync/patch.rs
arv: I added a test to enforce that in the follow up.

PR #326, opened 2021-03-08
tiny diffsync followups

phritz:

### DIFF
--- a/src/sync/pull.rs
+++ b/src/sync/pull.rs
@@ -283,6 +283,21 @@ pub async fn maybe_end_try_pull(
         .await
         .map_err(WriteSyncHeadError)?;
     dag_write.commit().await.map_err(CommitError)?;
+    let (old_last_mutation_id, old_cookie) = Commit::snapshot_meta_parts(&main_snapshot)
+        .map_err(|e| InternalProgrammerError(format!("{:?}", e)))?;
+    let (new_last_mutation_id, new_cookie) = Commit::snapshot_meta_parts(&sync_snapshot)
+        .map_err(|e| InternalProgrammerError(format!("{:?}", e)))?;
+    debug!(
+        lc.clone(),
+        "Successfully pulled new snapshot w/last_mutation_id={} (prev. {}),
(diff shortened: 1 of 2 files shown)


src/sync/pull.rs
```diff
@@ -281,6 +283,21 @@ pub async fn maybe_end_try_pull(
         .await
         .map_err(WriteSyncHeadError)?;
     dag_write.commit().await.map_err(CommitError)?;
+    let (old_last_mutation_id, old_cookie) = Commit::snapshot_meta_parts(&main_snapshot)
```
arv, line 286: Could these be wrapped in a `debug` check? Seems silly to do this work if we are not going to log them (which we do not do by default)

arv (commented):
LGTM

I always imagined we would include the `app_version`/`schemaVersion` in `push` too. It seems important for the server to know what version the client is at in a push request too.

phritz:
> I always imagined we would include the app_version/schemaVersion in push too. It seems important for the server to know what version the client is at in a push request too.

Recall for pull that pull_version indicates the version of the protocol the client is using which is hard-coded in repc, and schema_version indicates what kind of client view the app understands and is set by the customer's code. For push, we already have push_version to indicate the version of the protocol repc is using. I think what you are asking is that we include a schema_version in push as well, that would indicate... what? The schema of the mutator args? That makes sense I guess but want to be sure what you are suggesting.

arv:
I have been arguing for DO_FOO_v2 in the past. Lets just stick with that story until costumers keep shooting themselves in the foot.

> I think what you are asking is that we include a schema_version in push as well, that would indicate... what? The schema of the mutator args? 

Correct

phritz:
Well, too late :) https://github.com/rocicorp/repc/pull/327

src/sync/pull.rs
phritz: yes, done in https://github.com/rocicorp/repc/pull/327/commits/46d67d234a322306e94c2da01711b7099c56e3a6

PR #328, opened 2021-03-09
Return client ID in open

arv:
Towards https://github.com/rocicorp/replicache-sdk-js/issues/275

### DIFF
--- a/src/embed/dispatch.rs
+++ b/src/embed/dispatch.rs
@@ -113,7 +113,11 @@ async fn do_open(conns: &mut ConnMap, req: &Request) -> Response {
         return Err("db_name must be non-empty".into());
     }
     if conns.contains_key(&req.db_name[..]) {
-        return Ok("".into());
+        return Err(format!(
+            "Database \"{}\" has already been opened. Please close it before opening it again",
+            req.db_name
+        )
+        .into());
     }
 
     let open_req = serde_wasm_bindgen::from_value::<OpenRequest>(req.data.clone())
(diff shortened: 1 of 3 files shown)


tests/wasm.rs
```diff
@@ -839,76 +856,87 @@ async fn test_scan_with_index() {
         exclusive: bool,         // scan start_exclusive
         expected: Vec<(&str, &str, &str)>, // expected results of the scan
     ) {
-        for use_memstore in vec![true, false] {
```
arv, line 842: ignore whitespace diff using https://github.com/rocicorp/repc/pull/328/files?diff=split&w=1

This test was a bit flawed when I wanted to test it using MemStore. It really made no sense but it was failing now because closing the db removes the state when we use a MemStore

src/embed/dispatch.rs
```diff
@@ -113,7 +113,11 @@ async fn do_open(conns: &mut ConnMap, req: &Request) -> Response {
         return Err("db_name must be non-empty".into());
     }
     if conns.contains_key(&req.db_name[..]) {
-        return Ok("".into());
+        return Err(format!(
```
aboodman, line 116: I think we were relying on this behavior before... like it was allowed (api wise) to create multiple repliache instances with same name.

arv: Where are we relying on this? There are definitely things that do not work if we allow this. (subscriptions for sure but maybe other things too)

I also think it is a bad idea but if we want this to work we can make it work.

arv: If you call close on one of them then both are closed is another problem.

arv:
I'm going to land this as is. If we want to support this in the future we need to put in some work to make it work.

src/embed/dispatch.rs
aboodman: Hm, maybe some old sample app, or maybe just in my mental model.

TBC, this means that only one instance of `Replicache` can be constructed for each named cached. This is kind of a surprising restriction to me just from a JS API pov.

I see the bug you are talking about in `close` and I agree that if this isn't really working let's make it official.

But I won't be surprised if users trip over this and find it confusing / limiting.

PR #329, opened 2021-03-17
Add overlappingRequests flag to "open"

arv:
When overlappingRequests is true it is in an error if the pull/push
endpoint uses HTTP1/x. It needs to use H2 or H3.

fetch/rust_client.rs is not really working with this change since it
uses hyper 0.13 which does not support HTTP/2. There is a newer version
of hyper that supports h2 but hyper runs on top of tokio and we use std
async. We have a compat file but I could not get it to work correctly. I
need to spend more time on this if we ever want to support non web.

I tested the browser fetch path manually.

https://www.notion.so/Push-pull-scheduling-268fd33bbcf14c5cbced55a249f409aa#c959432fc61a4c4bb152aa61e3856f44

Towards https://github.com/rocicorp/replicache-sdk-js/issues/294

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -731,7 +731,7 @@ dependencies = [
  "futures-sink",
  "futures-task",
  "memchr",
- "pin-project 1.0.1",
+ "pin-project",
  "pin-utils",
  "proc-macro-hack",
  "proc-macro-nested",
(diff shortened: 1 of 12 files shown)


arv:
I'll do `orderedCookie` in a new PR

PR #330, opened 2021-03-19
Use durability relaxed for write transactions 

arv:
This uses `extern C` to expose js functions to Rust.

I ended up exposing more functions than `transaction` to reduce the
number of allocations and conversions needed between Rust and JS.

### DIFF
--- a/src/kv/idbstore.js
+++ b/src/kv/idbstore.js
@@ -1,0 +1,39 @@
+// @ts-check
+
+const RELAXED = {durability: "relaxed"};
+
+const OBJECT_STORE = "chunks";
+
+/**
+ * @param {IDBDatabase} idb
+ * @return {IDBTransaction}
+ */
+export function readTransaction(idb) {
+  return idb.transaction(OBJECT_STORE, "readwrite");
+}
+
+/**
+ * @param {IDBDatabase} idb
+ * @return {IDBTransaction}
+ */
+export function writeTransaction(idb) {
+  // TS does not have type defs for the third options argument yet.
+  // @ts-ignore Expected 1-2 arguments, but got 3.ts(2554)
+  return idb.transaction(OBJECT_STORE, "readwrite", RELAXED);
+}
+
+/**
+ * @param {IDBTransaction} tx
+ * @return {IDBObjectStore }
+ */
+export function objectStore(tx) {
+  return tx.objectStore(OBJECT_STORE);
+}
+
+/**
+ * @param {IDBDatabase} idb
+ * @return {IDBObjectStore}
+ */
+export function createObjectStore(idb) {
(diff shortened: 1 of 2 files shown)


src/kv/idbstore.rs
```diff
@@ -9,9 +9,20 @@ use futures::channel::oneshot;
 use futures::future::join_all;
 use std::collections::HashMap;
 use wasm_bindgen::closure::Closure;
+use wasm_bindgen::prelude::*;
 use wasm_bindgen::{JsCast, JsValue};
 use web_sys::{IdbDatabase, IdbFactory, IdbTransaction};
 
+#[wasm_bindgen(module = "/js/idb-store.js")]
```
aboodman, line 16: wow!

aboodman (approved):
Presumably we should just land this? 🤷‍♂️

arv:
@aboodman I ended up cleaning this up a little. Now there is no more need to allocate strings during db transactions.

PTAL

src/kv/idbstore.js
```diff
@@ -0,0 +1,39 @@
...
+const RELAXED = {durability: "relaxed"};
+
+const OBJECT_STORE = "chunks";
+
+/**
+ * @param {IDBDatabase} idb
+ * @return {IDBTransaction}
+ */
+export function readTransaction(idb) {
+  return idb.transaction(OBJECT_STORE, "readwrite");
```
aboodman, line 12: Two questions I guess:

(1) aren't const strings like `"chunks"` allocated statically anyway?
(2) do you want to make `"chunks"` and `"readwrite"` consistent?

aboodman (approved):

src/kv/idbstore.js
arv: > (1) aren't const strings like "chunks" allocated statically anyway?

Not across the wasm boundary.

> (2) do you want to make "chunks" and "readwrite" consistent?

Yes

arv: You are right that string literals are allocated statically in V8. I misunderstood you.

I wanted `OBJECT_STORE` external so that it is not repeated

arv: But more importantly the read transaction should use `"readonly"`! Ooops!

src/kv/idbstore.js
```diff
@@ -0,0 +1,38 @@
...
+
+const RELAXED = {durability: "relaxed"};
+const OBJECT_STORE = "chunks";
+
+/**
+ * @param {IDBDatabase} idb
+ * @return {IDBTransaction}
+ */
+export function readTransaction(idb) {
+  return idb.transaction(OBJECT_STORE, "readonly");
```
arv, line 11: Since "readonly" is only in one place I left it as a string literal.

I think V8 would inline it but 🤷

PR #337, opened 2021-04-05
Simplify release.sh slightly

arv:


aboodman:
LGTM, thanks.

tool/release.sh
```diff
@@ -29,18 +29,13 @@ fi
...
+    cp Cargo.toml Cargo.toml.bak
     perl -pi -e's/crate-type = .*/crate-type = ["cdylib"]/' Cargo.toml
 
     rm repc.zip
     rm -rf pkg
-    wasm-pack build --profiling -t web -- --no-default-features
-    brotli -f pkg/replicache_client.js pkg/replicache_client_bg.wasm
-    mv pkg debug
-    wasm-pack build --release -t web -- --no-default-features
-    brotli -f pkg/replicache_client.js pkg/replicache_client_bg.wasm
```
aboodman, line 40: I want the brotli line in there for two reasons:

1. I don't know how common it is for web servers to be setup to do this at runtime.
2. I want it to be clear what the true compressed size is at a glance.

Is there some downside to including it?

aboodman: 3. The runtime compression might not be as good. For example, Vercel's default brotli compression is 28% worse than what I get locally with the maximum settings.

I'd like to change all the samples that manually link to the object to use the brotli path for this reason.

arv: This answers my Q in the other PR. Let's change the npm package too then?

aboodman: Makes sense. Will do.

PR #338, opened 2021-04-05
Further optimize the wasm bundle with wasm-opt.

aboodman:
This shaves 10% off the compressed size and whatever it's doing, noticeably improves benchmarks too.

### DIFF
--- a/tool/release.sh
+++ b/tool/release.sh
@@ -34,8 +34,12 @@ fi
 
     rm repc.zip
     rm -rf pkg
-    wasm-pack build --profiling --target web --out-dir pkg/debug -- --no-default-features
+    wasm-pack build --debug --target web --out-dir pkg/debug -- --no-default-features
+    wasm-opt -O4 -o pkg/debug/replicache_client_bg.wasm pkg/debug/replicache_client_bg.wasm
+    brotli pkg/debug/replicache_client_bg.wasm
     wasm-pack build --release --target web --out-dir pkg/release -- --no-default-features
+    wasm-opt -O4 -o pkg/release/replicache_client_bg.wasm pkg/release/replicache_client_bg.wasm
+    brotli pkg/release/replicache_client_bg.wasm
     zip -r pkg pkg
     mv pkg.zip repc.zip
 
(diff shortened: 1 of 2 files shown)


aboodman:
@arv once you land your change I'll rebase this.

tool/release.sh
```diff
@@ -33,10 +33,11 @@ fi
 
     rm repc.zip
     rm -rf pkg
-    wasm-pack build --profiling -t web -- --no-default-features
+    wasm-pack build --debug -t web -- --no-default-features
     brotli -f pkg/replicache_client.js pkg/replicache_client_bg.wasm
     mv pkg debug
     wasm-pack build --release -t web -- --no-default-features
+    wasm-opt -O4 -o pkg/replicache_client_bg.wasm pkg/replicache_client_bg.wasm
```
arv, line 40: I thought that wasm-pack invoked wasm-opt already.

tool/release.sh
```diff
@@ -33,10 +33,11 @@ fi
 
     rm repc.zip
     rm -rf pkg
-    wasm-pack build --profiling -t web -- --no-default-features
+    wasm-pack build --debug -t web -- --no-default-features
```
arv, line 36: The debug build is 4MB or so. I thought you intentionally wanted to do a profiling build here?

arv (approved):

tool/release.sh
aboodman: I read that somewhere too, but running this takes off like 60KB.

tool/release.sh
aboodman: I did, but the two builds are the exact same size so it doesn't seem like profiling is doing what we think it is. The debug build compressed is only 50% bigger, so I think we should choose it instead afterall if we're going to have a debug build.

arv: profiling only includes the symbol table as far as I can tell... Otherwise it is the same.

I approve of using debug though but it does make profiling harder....

aboodman: > but it does make profiling harder

Oh right, duh.

The thing that's confusing to me is why the profiling and release builds are exactly the same size. IIRC, the release build does lack symbols when I look at in web inspector, while the profiling build has them. But that doesn't match the file sizes.

I'll have to look at this a little more.

arv:
Is there any action I need to take here?

aboodman:
No.

PR #339, opened 2021-04-08
Run wasm-opt explicitly as part of tools/release.

aboodman:
There are no prebuilt binaries for M1, so wasm-pack just skips it.
However you can manually install wasm-opt and run it, so let's do
that.

Also:
- Put the brotli compression back. I think it's good for us to
  put our best foot forward in the binary, so to speak.
- Change "profiling" to "debug". There's no difference in file
  size between profiling and release. I have no idea what it's
  supposed to change. I figure if we're going to have a debug
  build let's at lease maximize its utility. With wasm-opt and
  brotli, the difference is only ~50% (the debug build is 300k),
  so it seems reasonable to expose to users. Maybe it will be a
  better debugging experience.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,8 +87,11 @@ lto = true
 opt-level = 3
 panic = "abort"
 
+# It would be nice to use wasm-opt, but there are no prebuilt binaries for m1:
+# https://github.com/rustwasm/wasm-pack/issues/913
+# So we just run it separately in ./tool/release.sh.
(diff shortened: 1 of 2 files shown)


aboodman:
Closed the other to move the branch to my repo

arv (approved):
LGTM

What are the file sizes after this?

Why are we compressing with brotli again? The npm package does not contain them and I doubt many web servers are set up to magically use them if present.

PR #341, opened 2021-04-22
Compute changed keys in pull and mutate

arv:
Return the changed keys in mutation commits and maybe end pull.

Towards https://github.com/rocicorp/replicache/issues/63

### DIFF
--- a/src/db/read.rs
+++ b/src/db/read.rs
@@ -1,11 +1,9 @@
-use super::{
-    commit::{Commit, FromHashError},
-    index,
-};
+use super::commit::{Commit, FromHashError};
+use super::index;
 use crate::dag;
 use crate::prolly;
 use std::collections::hash_map::HashMap;
-use std::{convert::TryInto, string::FromUtf8Error};
+use std::convert::TryInto;
 
 #[derive(Debug)]
 pub enum Whence {
@@ -138,7 +136,6 @@ impl<'a> Read<'a> {
 #[derive(Debug)]
 pub enum ScanError {
     GetMapError(index::GetMapError),
-    InvalidUtf8(FromUtf8Error),
     ScanOptionsError(super::scan::ScanOptionsError),
     UnknownIndexName(String),
 }
--- a/src/db/read.rs
+++ b/src/db/read.rs
@@ -179,7 +179,7 @@ mod tests {
         )
         .await
         .unwrap();
-        w.commit(db::DEFAULT_HEAD_NAME, false).await.unwrap();
+        w.commit(db::DEFAULT_HEAD_NAME).await.unwrap();
 
         let dr = ds.read(LogContext::new()).await.unwrap();
         let r = OwnedRead::from_whence(Whence::Head(str!(db::DEFAULT_HEAD_NAME)), dr)
--- a/src/db/read.rs
+++ b/src/db/read.rs
@@ -1,9 +1,17 @@
-use super::commit::{Commit, FromHashError};
-use super::index;
+use async_std::sync::RwLock;
+
+use super::ScanOptions;
+use super::{
+    commit::{Commit, FromHashError},
+    index,
+};
 use crate::dag;
 use crate::prolly;
-use std::collections::hash_map::HashMap;
-use std::convert::TryInto;
+use std::{
+    collections::{hash_map::HashMap, BTreeSet},
+    sync::Arc,
+};
+use std::{convert::TryInto, string::FromUtf8Error};
 
 #[derive(Debug)]
 pub enum Whence {
@@ -12,10 +20,62 @@ pub enum Whence {
     Hash(String),
 }
 
+pub enum SubscriptionInnerError {
+    InvalidUtf8(std::string::FromUtf8Error),
+}
+
+pub struct SubscriptionInner {
+    key_set: BTreeSet<Vec<u8>>,
+    scans: Vec<ScanOptions>,
+}
+
+impl SubscriptionInner {
+    pub fn new() -> Self {
+        SubscriptionInner {
+            key_set: BTreeSet::new(),
+            scans: Vec::new(),
+        }
+    }
+
+    pub fn record_get(&mut self, key: Vec<u8>) {
+        self.key_set.insert(key);
+    }
+
+    pub fn record_has(&mut self, key: Vec<u8>) {
+        self.key_set.insert(key);
+    }
+
+    pub fn record_scan(&mut self, scan_options: ScanOptions) {
+        self.scans.push(scan_options);
+    }
+
+    // TODO(arv): Just make the error String::FromUtf8Error
+    pub fn touched_keys(&self) -> Result<Vec<String>, SubscriptionInnerError> {
+        // TODO: Consider draining the vec. We might not need this data more
+        // than once depending on usage.
+        use SubscriptionInnerError::*;
+
+        self.key_set
+            .iter()
+            .map(|b| String::from_utf8(b.to_vec()))
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(InvalidUtf8)
+    }
+
+    pub fn touched_scans(&self) -> Vec<ScanOptions> {
+        // TODO: Consider draining the vec. We might not need this data more
+        // than once depending on usage.
+        self.scans.clone()
+    }
+}
+
+pub type Subscription = Arc<RwLock<SubscriptionInner>>;
+
 pub struct OwnedRead<'a> {
     dag_read: dag::OwnedRead<'a>,
     map: prolly::Map,
     indexes: HashMap<String, index::Index>,
+    pub subscription: Option<Subscription>,
 }
 
 #[derive(Debug)]
@@ -30,6 +90,7 @@ impl<'a> OwnedRead<'a> {
     pub async fn from_whence(
         whence: Whence,
         dag_read: dag::OwnedRead<'a>,
+        subscription: Option<Subscription>,
     ) -> Result<OwnedRead<'a>, ReadCommitError> {
         let (_, basis, map) = read_commit(whence, &dag_read.read()).await?;
         let indexes = read_indexes(&basis);
@@ -37,11 +98,17 @@ impl<'a> OwnedRead<'a> {
             dag_read,
             map,
             indexes,
+            subscription: subscription.clone(),
         })
     }
 
     pub fn as_read(&'a self) -> Read<'a> {
-        Read::new(self.dag_read.read(), &self.map, &self.indexes)
+        Read::new(
+            self.dag_read.read(),
+            &self.map,
+            &self.indexes,
+            self.subscription.clone(),
+        )
     }
 }
 
@@ -85,6 +152,7 @@ pub struct Read<'a> {
     dag_read: dag::Read<'a>,
     map: &'a prolly::Map,
     indexes: &'a HashMap<String, index::Index>,
+    subscription: Option<Subscription>,
 }
 
 impl<'a> Read<'a> {
@@ -92,19 +160,29 @@ impl<'a> Read<'a> {
         dag_read: dag::Read<'a>,
         map: &'a prolly::Map,
         indexes: &'a HashMap<String, index::Index>,
+        subscription: Option<Subscription>,
     ) -> Read<'a> {
         Read {
             dag_read,
             map,
             indexes,
+            subscription,
         }
     }
 
-    pub fn has(&self, key: &[u8]) -> bool {
+    pub async fn has(&self, key: &[u8]) -> bool {
+        if let Some(sub) = self.subscription.clone() {
+            let mut subscription = sub.write().await;
+            subscription.record_has(key.to_vec());
+        }
         self.map.has(key)
     }
 
-    pub fn get(&self, key: &[u8]) -> Option<&[u8]> {
+    pub async fn get(&self, key: &[u8]) -> Option<&[u8]> {
+        if let Some(sub) = self.subscription.clone() {
+            let mut subscription = sub.write().await;
+            subscription.record_get(key.to_vec());
+        }
         self.map.get(key)
     }
 
@@ -114,6 +192,11 @@ impl<'a> Read<'a> {
         callback: impl Fn(super::scan::ScanResult<'_>),
     ) -> Result<(), ScanError> {
         use ScanError::*;
+
+        if let Some(subscription) = self.subscription.clone() {
+            let mut subscription = subscription.write().await;
+            subscription.record_scan(opts.clone());
+        }
 
         let opts_internal: super::scan::ScanOptionsInternal =
             opts.try_into().map_err(ScanError::ScanOptionsError)?;
@@ -133,9 +216,18 @@ impl<'a> Read<'a> {
     }
 }
 
+impl From<SubscriptionInnerError> for ScanError {
+    fn from(e: SubscriptionInnerError) -> Self {
+        match e {
+            SubscriptionInnerError::InvalidUtf8(e) => ScanError::InvalidUtf8(e),
+        }
+    }
+}
+
 #[derive(Debug)]
 pub enum ScanError {
     GetMapError(index::GetMapError),
+    InvalidUtf8(FromUtf8Error),
     ScanOptionsError(super::scan::ScanOptionsError),
     UnknownIndexName(String),
 }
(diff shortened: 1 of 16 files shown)


src/embed/connection.rs
```diff
@@ -309,12 +315,23 @@ async fn do_open_transaction<'a, 'b>(
 ) -> Result<OpenTransactionResponse, OpenTransactionError> {
     use OpenTransactionError::*;
 
+    // We create a Subscription object to keep track of what keys and scans were
+    // read in this subscribe-query transaction.
+    let subscription = if req.is_subscription {
+        let inner = SubscriptionInner::new();
```
arv, line 321: @phritz Turns out that I didn't need the global map. JS needs to maintain the subscription. The transaction owns the Subscription instance and it goes away when the transaction goes away.

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
     Hash(String),
 }
 
+pub enum SubscriptionInnerError {
```
arv, line 23: Better name and better file structure suggestions wanted

arv (commented):
I think this needs a few rounds of polish review but I think it is ready to start looking at

src/prolly/map.rs
```diff
@@ -116,6 +125,75 @@ impl Map {
...
+                        keys.push(String::from_utf8(key.clone())?);
+                    }
+                }
+            }
+        }
+        Ok(keys)
+    }
+
+    /// Returns the keys that are different between two maps.
+    pub fn diff<'a>(a: &'a Self, b: &'a Self) -> Result<Vec<String>, FromUtf8Error> {
```
arv, line 155: One optimization I didn't do is to create key ranges of this... we can do that later.

aboodman:
🎉

One question right off:

> We now return the touched keys in subscribe queries.

It seems like the JS would already have the information it needs here? It know what keys are being requested and the scan parameters.

arv:
> It seems like the JS would already have the information it needs here? It know what keys are being requested and the scan parameters.

That is true. My reason for moving this to Rust was code reuse in future non JS SDK but maybe that was a bad call? At one point I even wanted to keep track of the subscriptions in Rust but due to multiple tabs I moved the storage up to js.

aboodman:
On Thu, Apr 22, 2021 at 11:12 AM Erik Arvidsson ***@***.***>
wrote:

> It seems like the JS would already have the information it needs here? It
> know what keys are being requested and the scan parameters.
>
> That is true. My reason for moving this to Rust was code reuse in future
> non JS SDK but maybe that was a bad call? At one point I even wanted to
> keep track of the subscriptions in Rust but due to multiple tabs I moved
> the storage up to js.
>
It's up to you. If you find it easy enough to work in Rust, I don't care. I
just don't want hypothetical desire to work on other platforms in the
future to slow us down at all now. If you find yourself fighting with rust,
and it's easy to move it out to JS, do it.

arv:
> It's up to you. If you find it easy enough to work in Rust, I don't care. I
just don't want hypothetical desire to work on other platforms in the
future to slow us down at all now. If you find yourself fighting with rust,
and it's easy to move it out to JS, do it.

Doing it in JS to see how much complexity it removes.

src/embed/connection.rs
arv: After @aboodman hint I moved the whole subscription tracking to JS

src/db/write.rs
```diff
@@ -338,15 +344,39 @@ impl<'a> Write<'a> {
 
     // Return value is the hash of the new commit.
     #[allow(clippy::too_many_arguments)]
-    pub async fn commit(mut self, head_name: &str) -> Result<String, CommitError> {
+    pub async fn commit(
+        mut self,
+        head_name: &str,
+        generate_diffs: bool,
```
aboodman, line 350: I think we should use the name `ChangedKeySet` or `ChangeSet` or something similar to avoid confusion with the term `diff` which we already use elsewhere to mean something different.

src/db/write.rs
```diff
@@ -338,15 +344,39 @@ impl<'a> Write<'a> {
...
+    ) -> Result<(String, BTreeMap<String, Vec<String>>), CommitError> {
```
aboodman, line 351: Instead of `BTreeMap<...>` for the changes, why not define a named type somewhere. That would give an opp to document the meaning of the key and the value (is it index name and changed keys?)

aboodman, line 351: OOC, why BTreeMap vs HashMap?

src/embed/connection.rs
```diff
@@ -110,7 +115,7 @@ pub async fn process(store: dag::Store, rx: Receiver<Request>, client_id: String
     let mut recv = true;
 
     futures.push(Box::pin(connection_future(
-        &rx,
+        &receiver,
```
aboodman, line 118: You and your long names :).

src/embed/types.rs
```diff
@@ -46,13 +47,16 @@ pub struct OpenTransactionResponse {
...
+    #[serde(rename = "generateDiffs")]
+    pub generate_diffs: bool,
 }
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct CommitTransactionResponse {
     // Note: the field is named "ref" in go but "ref" is a reserved word in rust.
     #[serde(rename = "ref")]
     pub hash: String,
+    pub diffs: BTreeMap<String, Vec<String>>,
```
aboodman, line 59: Can you ensure this isn't significantly increasing the size of the js generated to serialize? I think it might be the first occurrence of BTreeMap going over this boundary.

src/prolly/map.rs
```diff
@@ -537,4 +616,100 @@ mod tests {
         )
         .await;
     }
+
+    macro_rules! map(
```
aboodman, line 620: I think it would be more clear to name this macro `prolly_map` since we have another macro we use in tests which is `hash_map!`.

src/sync/pull.rs
```diff
@@ -221,9 +226,15 @@ pub async fn maybe_end_try_pull(
         .await
         .map_err(GetMainHeadError)?
         .ok_or(MissingMainHead)?;
+
     let main_snapshot = Commit::base_snapshot(&main_head_hash, &dag_read)
         .await
         .map_err(NoBaseSnapshot)?;
+
+    let old_main_head_map = prolly::Map::load(main_snapshot.value_hash(), &dag_read)
```
aboodman, line 234: Delay loading this until `pending.is_empty()`? It's not needed in the case we are replaying.

src/sync/pull.rs
```diff
@@ -301,9 +319,16 @@ pub async fn maybe_end_try_pull(
             main_snapshot.value_hash()
         );
     }
+    let main_diff = diff.map_err(InvalidUtf8)?;
+    let mut diffs = BTreeMap::new();
```
aboodman, line 323: What about the index diffs? We need those too.

Seems like the code for computing the set of index diffs should be pulled out so it can be used by both CommitTransaction and Pull.

src/prolly/map.rs
```diff
@@ -116,6 +125,75 @@ impl Map {
...
+
+    // Returns the diff between the pending entries and the already flushed entries.
+    pub fn pending_diff(&self) -> Result<Vec<String>, FromUtf8Error> {
+        let mut keys = Vec::with_capacity(self.pending.len());
+        for (key, pending_val) in self.pending.iter() {
+            match pending_val {
+                Some(pending_val) => match self.base_get(key) {
+                    Some(base_val) => {
+                        if pending_val != base_val {
+                            keys.push(String::from_utf8(key.clone())?);
```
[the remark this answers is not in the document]
arv: Needs to be copied later so I copied it earlier since it made the code simpler. I can think a bit more about this to see if it makes sense to go `[u8]` all the way to JS

src/prolly/map.rs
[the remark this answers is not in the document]
arv: Yes, should be changed keys here and other places

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
...
+    }
+
+    pub fn touched_scans(&self) -> Vec<ScanOptions> {
+        // TODO: Consider draining the vec. We might not need this data more
+        // than once depending on usage.
+        self.scans.clone()
+    }
+}
+
+pub type Subscription = Arc<RwLock<SubscriptionInner>>;
```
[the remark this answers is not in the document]
arv: In theory we do not need locks because JS and Wasm is single threaded but I don't think Rust would allow that.

The following case needs "concurrent" access.

```ts
rep.subscribe(tx => 
  Promise.all([tx.get('a'), tx.get('b')])
);
```

I do not see how this could deadlock but it is always a good thing to keep in mind.

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
...
+
+pub struct SubscriptionInner {
+    key_set: BTreeSet<Vec<u8>>,
+    scans: Vec<ScanOptions>,
+}
+
+impl SubscriptionInner {
+    pub fn new() -> Self {
+        SubscriptionInner {
+            key_set: BTreeSet::new(),
```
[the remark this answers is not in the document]
arv: I don't think it is really needed. When I started I thought I could use linear traversal in a few places.

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
     Hash(String),
 }
 
+pub enum SubscriptionInnerError {
+    InvalidUtf8(std::string::FromUtf8Error),
+}
+
+pub struct SubscriptionInner {
+    key_set: BTreeSet<Vec<u8>>,
```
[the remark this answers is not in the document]
arv: or watched_keys

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
...
+    scans: Vec<ScanOptions>,
```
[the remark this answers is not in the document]
arv: Thanks. I had a custom struct for a while but it ended up being mostly the same in the end

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
...
+            key_set: BTreeSet::new(),
+            scans: Vec::new(),
+        }
+    }
+
+    pub fn record_get(&mut self, key: Vec<u8>) {
+        self.key_set.insert(key);
+    }
+
+    pub fn record_has(&mut self, key: Vec<u8>) {
```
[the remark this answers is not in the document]
arv: Sure. It was mostly for self documenting code.

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
...
+        self.key_set.insert(key);
+    }
+
+    pub fn record_scan(&mut self, scan_options: ScanOptions) {
+        self.scans.push(scan_options);
+    }
+
+    // TODO(arv): Just make the error String::FromUtf8Error
```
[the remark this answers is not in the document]
arv: I think I'll try to keep all of these as `Vec<u8>` or `&[u8]` in a followup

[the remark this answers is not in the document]
arv: The comment was that it was silly to wrap a single error in a new error struct.

src/db/read.rs
```diff
@@ -12,10 +20,62 @@ pub enum Whence {
...
+    }
+}
+
+pub type Subscription = Arc<RwLock<SubscriptionInner>>;
+
 pub struct OwnedRead<'a> {
     dag_read: dag::OwnedRead<'a>,
     map: prolly::Map,
     indexes: HashMap<String, index::Index>,
+    pub subscription: Option<Subscription>,
```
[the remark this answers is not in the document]
arv: +1

src/db/read.rs
```diff
@@ -1,9 +1,11 @@
-use super::commit::{Commit, FromHashError};
-use super::index;
+use super::{
```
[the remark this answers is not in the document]
arv: Strange. I don't see a yellow squiggly

src/db/read.rs
```diff
@@ -136,6 +138,7 @@ impl<'a> Read<'a> {
 #[derive(Debug)]
 pub enum ScanError {
     GetMapError(index::GetMapError),
+    InvalidUtf8(FromUtf8Error),
```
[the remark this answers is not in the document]
arv: I'll check

src/db/write.rs
arv: I agree.

src/db/write.rs
arv: +1

src/embed/connection.rs
arv: I found the var `tx` too confusing given that we use that all over for transaction.

aboodman: +1.

src/embed/types.rs
arv: It is the first, I would also be fine going through a `Vec<(String, String)>`. 

Let me try to figure out how it affects size.

src/prolly/map.rs
arv: +1

src/sync/pull.rs
arv: Will look into it.

src/sync/pull.rs
arv: oops

arv: We do not seem to update the index at all in this code path? Maybe it is done somewhere else?

arv: Figured out what is going on. We update the index when applying the patch and then again when we replay the mutations. I will have to load the indexes from before the patch and the indexes from after all the replays and diff all of them.

src/db/write.rs
arv: ChangedKeysMap

src/sync/pull.rs
aboodman: Yeah the indexes must be recomputed to match the state before each replay begins (when the mutation runs, it must see the indexes in the correct state to correspond to that cache state).

> I will have to load the indexes from before the patch and the indexes from after all the replays and diff all of them.

Yep!

src/db/write.rs
arv: I thought I answered this already. I thought that I might take advantage of ordered sets/vectors at some point. Never did that. Switching to HashMap for now.

src/sync/pull.rs
```diff
@@ -1203,6 +1297,7 @@ mod tests {
                 intervening_sync: false,
                 exp_replay_ids: vec![],
                 exp_err: None,
+                exp_changed_keys: ChangedKeysMap::new(),
```
arv, line 1300: @phritz It seems like this test are all about nop mutations and pulls? I was looking for a rust/wasm test that does pull with changes but I guess we didn't need to test this before?

src/embed/types.rs
arv:
```
With Map:
replicache_client_bg.wasm: 773140
replicache_client_bg.wasm.br: 209530

With Entries:
replicache_client_bg.wasm: 774323
replicache_client_bg.wasm.br: 209972

Diff 
replicache_client_bg.wasm: 1183
replicache_client_bg.wasm.br: 442
```

Using a `HashMap<String, Vec<String>>` is more efficient than `Vec<(String, Vec<String>)>`

src/sync/pull.rs
```diff
@@ -1343,4 +1459,175 @@ mod tests {
             };
         }
     }
+
+    #[async_std::test]
+    async fn test_changed_keys() {
```
arv, line 1464: @phritz I cannot figure out how to write this test.

I want to have the snapshot contain a non empty map and an index. Then I want to run begin and end pull with a patch that would update the index. Then I want to check the diff.

Maybe the test is fine and the implementation is wrong? I just can't figure this out and I have spent way too much time on this test already.

Please help

src/prolly/map.rs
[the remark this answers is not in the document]
arv: Going to do this in another PR

arv (commented):
Ready for realz. PTAL

src/db/read.rs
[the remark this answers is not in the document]
phritz: At least some of the proliferation of locks in the rust code is due to rust not allowing inner mutability without them. Where in another language we wouldn't bother putting a lock around a thing that we know is not going to be accessed concurrently, rust makes us.

src/sync/pull.rs
```diff
@@ -1343,4 +1453,223 @@ mod tests {
...
+                sync_head: pull_result.sync_head.clone(),
+            };
+            let result = maybe_end_try_pull(&store, LogContext::new(), req)
+                .await
+                .unwrap();
+
+            assert_eq!(result.changed_keys, expected_changed_keys_map,);
+        }
+
+        test(
```
phritz, line 1549: Think we should also have a test with Operation::Clear to be safe.

src/sync/types.rs
```diff
@@ -30,6 +32,8 @@ pub struct MaybeEndTryPullResponse {
     pub replay_mutations: Vec<ReplayMutation>,
     #[serde(rename = "syncHead")]
     pub sync_head: String,
+    #[serde(rename = "changedKeys")]
+    pub changed_keys: ChangedKeysMap,
```
phritz, line 36: Maybe say here that changed_keys will only be filled if A) the caller requested it and B) on the last MaybeEndTryPull call, when there are no more to replay.

phritz (commented):
I gave this a careful reading and it looks great! Only a couple minor comments.

PR #342, opened 2021-05-03
Simplify IdbStore

arv:
By moving more code to JS

### DIFF
--- a/src/kv/idbstore.rs
+++ b/src/kv/idbstore.rs
@@ -55,12 +55,6 @@ impl From<JsValue> for StoreError {
     }
 }
 
-impl From<futures::channel::oneshot::Canceled> for StoreError {
-    fn from(_e: futures::channel::oneshot::Canceled) -> StoreError {
-        StoreError::Str("oneshot cancelled".into())
-    }
-}
-
 pub struct IdbStore {
     // We would like:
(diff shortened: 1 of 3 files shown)


src/kv/idbstore.rs
```diff
@@ -106,84 +112,37 @@ pub struct IdbStore {
...
-        request.set_onupgradeneeded(Some(onupgradeneeded.as_ref().unchecked_ref()));
-        receiver.await?;
-
-        let db: IdbDatabase = request.result()?.into();
-        let db_copy = db.clone();
-        let onversionchange = Closure::once(move |_event: web_sys::IdbVersionChangeEvent| {
-            db_copy.close();
-        });
-        db.set_onversionchange(Some(onversionchange.as_ref().unchecked_ref()));
-        onversionchange.forget();
```
arv, line 140: I started all of this work because of this line. I was feeling anxiety about this and thought that it might have been the reason for the slowdown. It was not.

But after going down this route I felt that this PR was an approvement even if it does not fix the issue.

aboodman (approved):
Really nice. Were you waiting on me here to LG this for some reason?

arv:
Not really waiting on anything. It was not not blocking in any way

PR #343, opened 2021-05-04
Use Vec<8> instead of string for ChangedKeys

arv:
All the way to the RPC barrier, at which point we turn the Vec<u8> to a
String.


arv:
I could not get this to share the underlying `[u8]` and using `RC` seems overkill for this.

I think if I refactor `prolly::Map::flush` to return the pending changed keys maybe I can get the ownership working.

PR #347, opened 2021-05-08
first set of DagMap building blocks

phritz:
progress towards https://github.com/rocicorp/rocicorp-internal/issues/17


aboodman:
Exciting!

phritz:
> Exciting!

Yes, it's basic but it's a start!

src/btree/node.rs
```diff
@@ -0,0 +1,289 @@
...
+#[derive(Debug)]
+pub enum Node<'a> {
+    Internal(fb::InternalNode<'a>),
+    Leaf(fb::InternalNode<'a>),
+    Data(fb::DataNode<'a>),
+}
+
+impl Node<'_> {
+    // key_iter returns an Iterator that iterates the keys contained in the node.
+    pub fn key_iter<'b, 'a: 'b>(&'a self) -> Box<dyn Iterator<Item = &'a [u8]> + 'b> {
```
phritz, line 32: Someone is going to suggest why not just have an iter over both key and value and ignore the value when it's not needed. Good suggestion! I think when I started this I mistakenly thought for some reason I shouldn't iterate the values, like in case they are large. But since the value is just a reference to something it's not going to be large itself. I'll make that change in the next PR.

phritz: Oh, I remember now. The values in data nodes are [u8] but the values internal nodes are chunk hashes aka &strs, so not obvious how to share an interator.

PR #348, opened 2021-05-08
Node::try_from.

sayrer:
I got this to work with named lifetimes. The errors for anonymous lifetimes were _really_ confusing.


phritz (approved):
Thanks!

phritz:
Looks like you just need to cargo fmt... (Going afk, one of us will merge soonish, thanks again!)

sayrer:
I got a bunch of files changed running fmt with cargo 1.52. I'd suggest cribbing this script from Rustls, so the format operation is the same for everyone: https://github.com/ctz/rustls/blob/main/admin/fmt

aboodman:
Actually the fmt seems addressed so will merge now.

PR #353, opened 2021-05-19
Made open, close and get_root exported functions

arv:
These can be directly invoked from JS now

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -1,11 +1,16 @@
-use std::sync::Once;
+use crate::{
+    dag,
+    db::{self, GetRootError},
+    embed,
+    kv::memstore::MemStore,
+};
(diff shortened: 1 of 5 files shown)


src/wasm.rs
```diff
@@ -38,3 +43,93 @@ fn init_panic_hook() {
     console_error_panic_hook::set_once();
     init_console_log();
 }
+
+// thread_local! {
+//     static KV_STORES: HashMap<String, Box<dyn Store>> = Default::default();
+// }
+
+thread_local! {
+    pub static KV_STORES: RefCell<HashMap<String, Arc<dyn Store>>> = Default::default();
```
arv, line 52: In the old world the state was kept in an async function with an iloop. Now it is a global and the async iloop is gone. The calls just originate from JS.

Issue #352, opened 2021-05-19
Return JsErrors in Wasm


arv:
We are losing information when we just stringify the error message.

## Proposal

Have all our errors implement a trait that has:

```rust
trait IntoJsError {
  fn maybe_into_js_error(&self) -> Option<js_sys::JsError>;
}
```

Then we can extract the inner error if there was one.

There is also the [ES stage 3 proposal for error cause](https://github.com/tc39/proposal-error-cause) which we could use if we wanted. If we decide to do something like that we can create a JsError at the top wasm exit point and attach the underlying error, if any.

arv:
@sayrer is there a common trait like `TryInto` that returns `Option` instead of `Result`? Thanks.

phritz:
Would be super useful to get line numbers if possible, or the current function, or similar. Would have helped with https://github.com/rocicorp/repc/issues/351 bc there are several places in the code where this could be coming from.

sayrer:
> @sayrer is there a common trait like `TryInto` that returns `Option` instead of `Result`? Thanks.

I _think_ what you want is this crate: https://crates.io/crates/anyhow. If that's not right, consider going back to matching on a crate enum, like Rustls does here: https://github.com/ctz/rustls/blob/beef0be1658cd457d4f47039abf4f4bb37d59a78/rustls/src/error.rs#L36

arv:
@aboodman I know you have a string opinion about this.

What I want is to return the original JS errors in `dispatch`. For example if IDB throws I would like that exception to be returned.

aboodman:
But don't you also sort of want the "stack" from Rust too? Ideally?

arv:
Wasm is already on the stack. But that would generally not contain meaningful names unless we are doing a profile build. At the moment we do not have a reason to have a non profile build. See https://github.com/rocicorp/replicache/issues/356

Issue #354, opened 2021-05-19
provide custom debug output for our error types


phritz:
In https://github.com/rocicorp/repc/issues/351 we can see a DbReadError and a PullFailed error. These are variants of OpenTransactionError and BeginTryPullError respectively. However when printed the name of the enum (OTE and BTPE) is not printed, just the variant (DRE and PF) is printed. This is because we derive Debug for all our errors, and the default implementation is to print the variant but not the enum name: 
```
When derived for ... enums, it will use the name of the variant 
```
-- https://doc.rust-lang.org/beta/std/fmt/trait.Debug.html

We need to implement Debug for our errors such that the enum name is printed along with the variant and anything it contains.

arv:
It is not clear to me if we need to print the whole call stack of errors?

PR #355, opened 2021-05-19
Throw an error instead of a string

arv:
At the top level exit point from wasm throw an Error instead of a
string.

Towards #352


arv:
This does not yet unpack errors that originated from JS in the first place.

phritz:
@arv will this address https://github.com/rocicorp/repc/issues/354, at least on the js side?

arv:
It will throw errors so we do get the stack trace outside of wasm.

The goal of #352 was to return the original error if it originated in JS (for example if there was an IDB error)

arv:
Argh,clippy. Will look tomorrow

PR #359, opened 2021-05-22
Ignore the response body in push

### LINKED ISSUE #349: We should ignore any body in the response from the push endpoint
We don't actually care what the response to push is, as long as it's `HTTP 200`.

But right now we validate that it is `{}` (this dates from a time when there was data in there we cared about).

Several users (incl me :)) have hit this.

arv:
Fixes #349


aboodman (approved):
🙏

PR #365, opened 2021-06-17
Add a JS backed Puller/Pusher implementation

arv:
This creates implementations of Puller and Pusher that are backed by a
JS Function. This function gets passed in to BeginTryPull and TryPush.


Cargo.toml
```diff
@@ -27,7 +27,7 @@ maplit = "1.0.2"
 sha2 = "0.8.1"
 serde = "1.0.116"
 serde_json = "1.0"
-serde-wasm-bindgen = "0.1.3"
+serde-wasm-bindgen = "0.3.0"
```
arv, line 30: This update was needed to pick a fix for internally tagged enums as well as support for `serialize_maps_as_objects(true)`

arv:
FetchPusher/FetchPuller and browser_client.rs fetch are now dead code. I'll remove later

PR #367, opened 2021-08-04
chore: Automating release of zip file

arv:
Use an action to build the zip file and add it to a release.


aboodman:
Wow! Awesome, thanks.

